PR #8, opened 2019-07-24
TC1 feature with hardcoded value was added

AvocadoGag:


cypress/integration/login.js
```diff
@@ -21,7 +21,27 @@ describe('Sourcebooks login', function() {
...
+        cy.get('[type="submit"]').click();
+
+        cy.url().should('include', '/time-logging');
+        cy.get('.page__title').contains('Timesheets');
+        cy.get('.calendar').should('be.visible');
+        cy.get('.tile.form').should('be.visible');
+        cy.get('.user-info__title').contains('Vitalij Janukevic');
+        cy.get('.main-nav').find('li').should('have.length', 1);
+        let currentDate = new Date();
+        cy.get('.calendar__date').contains(currentDate.getDate());
```
nikolaj-tolkaciov: I would recommend merging this test case with a previous one as almost all the steps are the same except one added line. That way you would save 10 seconds of execution time

nikolaj-tolkaciov: This ".calendar__date" should not work :) this is a cypress feature that this works, as calendar__date class have multiple occurrences, there is a specific locator for today, please use that one

cypress/integration/login.js
```diff
@@ -21,7 +21,9 @@ describe('Sourcebooks login', function() {
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Vitalij Janukevic');
         cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.calendar--today').contains(new Date.getDate());
     })
+
```
nikolaj-tolkaciov: There is no need to add a gap between last test case and end of test cases

PR #12, opened 2019-07-24
added check if current date is selected correctly

RMNT:


cypress/integration/login.js
```diff
@@ -21,7 +21,18 @@ describe('Sourcebooks login', function() {
...
+
+    it('Should be able to check if today date is selected', function() {
+        cy.visit('/');
+        cy.get('[id="loginForm.userId"]').click({force:true});
+        cy.get('[aria-label="Raminta Urbonavičiūtė"]').click();
+        cy.get('[id="loginForm.role"]').click({force:true});
+        cy.get('[aria-label="User"]').click();
+        cy.get('[type="submit"]').click();
+        var currentDate = new Date();
+        cy.get('[class="calendar__day calendar--today calendar--selected"]').contains(currentDate.getDate())
```
nikolaj-tolkaciov: This test case seems like a same as above but with extra assertion. In my opinion it would make sense to combine them, and save extra 10 seconds

PR #15, opened 2019-07-24
Check if today's date is marked on calendar.

bertarejabut:


cypress/integration/login.js
```diff
@@ -21,7 +21,10 @@ describe('Sourcebooks login', function() {
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Berta Reja Butvilaite');
         cy.get('.main-nav').find('li').should('have.length', 1);
     })
+    it('Should be able to see today\'s date marked on calendar', function () {
+        cy.get('.calendar--today').contains(new Date().getDate());
```
nikolaj-tolkaciov: This test case depends on the previous one, if previous one is deleted or another test case is inserted in between this one is likely to fail

PR #21, opened 2019-07-24
Today date

Cheekiuz:


cypress/integration/login.js
```diff
@@ -21,7 +21,11 @@ describe('Sourcebooks login', function() {
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Ugne Lekstutiene');
         cy.get('.main-nav').find('li').should('have.length', 1);
     })
+
+    it('Date today', function()  {
+        cy.get('.calendar--today').contains('24');
```
nikolaj-tolkaciov: This test case is very depended on a previous one, if previous one will change or another one is added in between this test case might start failing

cypress/integration/login.js
```diff
@@ -12,16 +12,20 @@ describe('Sourcebooks login', function() {
...
         cy.visit('/');
         cy.get('[id="loginForm.userId"]').click({force:true});
-        cy.get('[aria-label="Demo User"]').click();
+        cy.get('[aria-label="Ugne Lekstutiene"]').click();
         cy.get('[id="loginForm.role"]').click({force:true});
         cy.get('[aria-label="User"]').click();
         cy.get('[type="submit"]').click();
 
         cy.url().should('include', '/time-logging');
+        cy.get('.calendar--today').contains(new Date().getDate());
```
nikolaj-tolkaciov: It is best to move this after line: 
cypress is flexible on this one and not failing, but other frameworks might fail because it is not yet loaded

cypress/integration/login.js
```diff
@@ -12,16 +12,20 @@ describe('Sourcebooks login', function() {
...
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Ugne Lekstutiene');
         cy.get('.main-nav').find('li').should('have.length', 1);
-    })
+        
+    })   
+
+
```
nikolaj-tolkaciov: There is some unnecessary gaps

cypress/integration/login.js
```diff
@@ -12,16 +12,19 @@ describe('Sourcebooks login', function() {
...
         cy.get('.calendar').should('be.visible')
+        cy.get('.calendar--today').contains(new Date().getDate());
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Ugne Lekstutiene');
         cy.get('.main-nav').find('li').should('have.length', 1);
-    })
+        
+    })   
+
```
linukkis: Why the double-line gap here?

cypress/integration/login.js
```diff
@@ -12,16 +12,19 @@ describe('Sourcebooks login', function() {
...
+        cy.get('[aria-label="Ugne Lekstutiene"]').click();
         cy.get('[id="loginForm.role"]').click({force:true});
         cy.get('[aria-label="User"]').click();
         cy.get('[type="submit"]').click();
-
-        cy.url().should('include', '/time-logging');
+        cy.url().should('include', '/time-logging');        
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
+        cy.get('.calendar--today').contains(new Date().getDate());
```
linukkis: This dynamic solution is great! However, this should probably be extracted into a separate test case or if not, the name of the test case should reflect this day assertion, e.g. "... and check if today's date is selected"

PR #22, opened 2019-07-24
verify current date

JevgenijaM:
Change Demo User to actual user name (Jevgenija Malomuž) and verify if current date is 24

### DIFF
--- a/cypress/integration/login.js
+++ b/cypress/integration/login.js
@@ -18,8 +18,8 @@ describe('Sourcebooks login', function() {
         cy.get('[type="submit"]').click();
 
         cy.url().should('include', '/time-logging');
-        cy.get('.calendar--today').contains('24')
         cy.get('.page__title').contains('Timesheets')
+        cy.get('.calendar--today').contains(new Date().getDate());
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
         cy.get('.user-info__title').contains('Jevgenija Malomuž');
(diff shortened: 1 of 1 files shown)


cypress/integration/login.js
```diff
@@ -12,16 +12,17 @@ describe('Sourcebooks login', function() {
...
         cy.visit('/');
         cy.get('[id="loginForm.userId"]').click({force:true});
-        cy.get('[aria-label="Demo User"]').click();
+        cy.get('[aria-label="Jevgenija Malomuž"]').click();
         cy.get('[id="loginForm.role"]').click({force:true});
         cy.get('[aria-label="User"]').click();
         cy.get('[type="submit"]').click();
 
         cy.url().should('include', '/time-logging');
+        cy.get('.calendar--today').contains('24')
```
nikolaj-tolkaciov: I would recommend to add this assert after line " cy.get('.calendar').should('be.visible')" because after that you will be able to see calendar. Cypress is smart enough to wait for element to appear, but other frameworks are not

cypress/integration/login.js
```diff
@@ -12,16 +12,17 @@ describe('Sourcebooks login', function() {
...
         cy.get('[id="loginForm.userId"]').click({force:true});
-        cy.get('[aria-label="Demo User"]').click();
+        cy.get('[aria-label="Jevgenija Malomuž"]').click();
         cy.get('[id="loginForm.role"]').click({force:true});
         cy.get('[aria-label="User"]').click();
         cy.get('[type="submit"]').click();
 
         cy.url().should('include', '/time-logging');
         cy.get('.page__title').contains('Timesheets')
+        cy.get('.calendar--today').contains(new Date().getDate());
```
nikolaj-tolkaciov: Please move this line after line: 
cypress is flexible and not failing in this case but other frameworks are not waiting for element to be visible

PR #23, opened 2019-07-24
Changed that connects with my name and with role Team Lead.  Checked …

ieva-st:
…that todays day is 24th.

### DIFF
--- a/cypress/integration/login.js
+++ b/cypress/integration/login.js
@@ -23,6 +23,6 @@ describe('Sourcebooks login', function() {
         cy.get('.tile.form').should('be.visible')
         cy.get('.user-info__title').contains('Ieva Stakneviciute');
         cy.get('.main-nav').find('li').should('have.length', 2);
-        cy.get('.calendar__day.calendar--today.calendar--selected').contains('24');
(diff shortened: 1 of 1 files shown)


cypress/integration/login.js
```diff
@@ -12,16 +12,17 @@ describe('Sourcebooks login', function() {
...
 
         cy.url().should('include', '/time-logging');
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
-        cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.user-info__title').contains('Ieva Stakneviciute');
+        cy.get('.main-nav').find('li').should('have.length', 2);
+        cy.get('.calendar__day.calendar--today.calendar--selected').contains('24');
```
nikolaj-tolkaciov: This locator is too long, it can be much shorter to find specific element. Also because this locator contains ".calendar--selected" it might be that if you select another date in the calendar it will fail to assert today's date

PR #24, opened 2019-07-24
Changed user to Benediktas Gilis, added line to check if the date is right

Lansing12:


cypress/integration/login.js
```diff
@@ -21,7 +21,8 @@ describe('Sourcebooks login', function() {
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Benediktas Gilis');
         cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.calendar').find('.calendar--selected').contains('24');
```
nikolaj-tolkaciov: Date is hard-coded and this test case will fail tomorrow
also ".calendar--selected" is for selected date but it is not necessary today, it could be last edited date

PR #25, opened 2019-07-24
Nadezda zarikova/feature/tc1 verify current date

NadezdaZarikova:


cypress/integration/login.js
```diff
@@ -21,7 +21,8 @@ describe('Sourcebooks login', function() {
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Nadežda Žarikova');
         cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.calendar__day.calendar--today.calendar--selected').contains(24);
```
nikolaj-tolkaciov: This locator is too long, it can be much shorter to find specific element. Also because this locator contains ".calendar--selected" it might be that if you select another date in the calendar it will fail to assert today's date

cypress/integration/login.js
```diff
@@ -21,7 +21,8 @@ describe('Sourcebooks login', function() {
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Nadežda Žarikova');
         cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.calendar__date').contains(new Date().getDate());
```
linukkis: Great function! However, don't user "calendar__date", because this class can be found in every single day's cell. Try using "calendar--today" instead :)

PR #27, opened 2019-07-24
added check if current date is selected correctly

RMNT:
some changes

### DIFF
--- a/cypress/integration/login.js
+++ b/cypress/integration/login.js
@@ -25,6 +25,6 @@ describe('Sourcebooks login', function() {
         cy.get('.main-nav').find('li').should('have.length', 1);
 
         var currentDate = new Date();
-        cy.get('[class="calendar__day calendar--today calendar--selected"]').contains(currentDate.getDate())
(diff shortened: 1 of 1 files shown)


cypress/integration/login.js
```diff
@@ -21,7 +21,10 @@ describe('Sourcebooks login', function() {
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Raminta Urbonaviciute');
         cy.get('.main-nav').find('li').should('have.length', 1);
+
+        var currentDate = new Date();
+        cy.get('[class="calendar__day calendar--today calendar--selected"]').contains(currentDate.getDate())
```
nikolaj-tolkaciov: this locator is too long and depending on many classes. Also it depends on selected date, and it could be other then today.

Also if you use class then it can be just cy.get('.calendar__day.calendar--today.calendar--selected')

PR #28, opened 2019-07-24
Dovile Martinkute/develop

mdovile:


cypress/integration/login.js
```diff
@@ -21,7 +21,8 @@ describe('Sourcebooks login', function() {
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Dovile Martinkute');
         cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.calendar__date').contains('24');
```
Imlus: This test won't work tomorrow as tomorrow will be 25th

Imlus: .calendar__date will not check the correct day :)

PR #59, opened 2019-07-25
Refactor to use page objects.

bertarejabut:


cypress/objects/loginPage.js
```diff
@@ -0,0 +1,26 @@
+class LoginPage {   //klasiu pav is didziosios
+    visit() {
+        cy.visit('/');  //homepage
+    }
+    getValidationError() {
+        return cy.get('.Select.not-valid');
+    }
+    getSubmit() {
```
linukkis: You could add "button" to the end, this way it's always clearer what kind of element we're interacting with, in this case a button :)

`getSubmitButton()`

cypress/objects/loginPage.js
```diff
@@ -0,0 +1,26 @@
...
+    getUserID() {
+        return cy.get('[id="loginForm.userId"]');
+    }
+    getUserOption(userName) {
+        return cy.get(`[aria-label="${userName}"]`);
+    }
+    getRole() {
+        return cy.get('[id="loginForm.role"]');
+    }
+    getRoleOption() {
```
linukkis: Here you could do exactly what you did with `getUserOption` and insert a parameter like `roleName`, so that it can be reused not only with the 'User' role, but with others as well :)

cypress/integration/login.js
```diff
@@ -1,30 +1,28 @@
+import LoginPage from '../objects/loginPage';   //.. - objects up
+import TimeLogging from '../objects/timeLogging';
 describe('Sourcebooks login', function() {
```
linukkis: Feels like this part could use an empty line to separate the imports from the describe block :)

cypress/objects/timeLogging.js
```diff
@@ -0,0 +1,19 @@
+class TimeLogging {   //klasiu pav is didziosios
```
linukkis: Oh also one more thing: it's great to keep comments, we all do, but when working on actual projects in the future try not to commit such comments that you make for yourself, rather write stuff like this down somewhere else or remove it when you're committing. But it's alright here in our lectures :)

bertarejabut:
This new pull request includes TC9 as well as TC3: Admin creates new task.

cypress/objects/taskPage.js
```diff
@@ -0,0 +1,36 @@
+class Tasks {
+    getTasks() {
+        return cy.get('[href="/tasks"]');
+    }
+    getButton() {
+        return cy.get('.btn');
```
Imlus: need more accurate name than generic getButton, I need to understand what I get here

cypress/objects/taskPage.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+    getFormName() {
+        return cy.get('[id="taskDetailsForm.name"]');
+    }
+    getFieldTextArea() {
+        return cy.get('.field__textarea');
+    }
+    getValue() {
+        return cy.get('.Select-value-label');
```
Imlus: need more accurate name than generic getValue, I need to understand what I get here

cypress/integration/task.js
```diff
@@ -0,0 +1,30 @@
...
+        tasks.getTasks().click();
+        tasks.getButton().click();
+        var autoName = 'Auto name_' + Math.random().toString(36).substr(2, 9);
+        tasks.getFormName().type(autoName);
+        tasks.getFieldTextArea().type('auto description');
+        tasks.getValue().click();
+        tasks.getValueLabel("Yes").click();
+        tasks.getFormRate().click();
+        tasks.getFormRate().clear();
+        tasks.getFormRate().type(1.234);
```
Imlus: can't we type text in a field without clicking and clearing the values?

cypress/objects/taskPage.js
```diff
@@ -0,0 +1,36 @@
...
+        return cy.url();
+    }
+    getSmallFieldText() {
+        return cy.get('.field__text--small');
+    }
+    getTitleName(name) {
+        return cy.get(`[title="${name}"]`);
+    }
+}
+export default Tasks;
```
linukkis: I've just noticed you exported this as "Tasks" instead of "TaskPage", which would immediately suggest what the class is about. Don't you think? :) You've already done this for "LoginPage", but not for "TimeLogging" though. This is not critical but to be kept in mind :)

cypress/objects/taskPage.js
```diff
@@ -11,10 +11,10 @@ class Tasks {
     getFieldTextArea() {
         return cy.get('.field__textarea');
     }
-    getValue() {
+    getBillToClient() {
```
linukkis: You could add "InputField" at the end, because the way it is now it's difficult to tell whether this is an input field, a label, a dropdown etc. I'd go for "getBillToClientInputField"

cypress/objects/taskPage.js
```diff
@@ -0,0 +1,36 @@
...
+    getFormRate() {
+        return cy.get('[id="taskDetailsForm.rate"]');
+    }
+    getSubmitButton() {
+        return cy.get('[type="submit"]');
+    }
+    getUrl() {
+        return cy.url();
+    }
+    getSmallFieldText() {
```
linukkis: I spent quite a bit figuring out what "SmallFieldText" is :D Try to name your functions logically according to what they refer to and not the name of the CSS class you're getting. For example, you could change `getSmallFieldText()` into `getTaskNameInputField()` or something like that. Makes it a bit clearer what we're talking about, doesn't it? :)

cypress/objects/taskPage.js
```diff
@@ -11,7 +11,7 @@ class Tasks {
     getFieldTextArea() {
```
linukkis: Just like with "small field text" this is not descriptive, should probably be changed to something like "getTaskDescriptionTextArea", agree? You can fix this in this pull-request or some other, but otherwise I'm giving this one a go, amazing job :) 👍

PR #60, opened 2019-07-25
Migle budryte/feature/tc9 refactor to use page object model

MigleBudryte:


cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,39 @@
...
+
+    getLogHoursForm(){
+        return cy.get('.tile.form')
+    }
+
+    getUserInfoTitle(){
+        return  cy.get('.user-info__title')
+    }
+
+    findMainNavTabs(){
```
linukkis: I think this could be `getMainNavTabs`, since what you actually do is get these tabs, even if the method is called "find" :)

cypress/integration/login.js
```diff
@@ -1,11 +1,17 @@
+import LoginPage from '../objects/loginPage'
+import TimeLoggingPage from '../objects/timeLoggingPage'
+
 describe('Sourcebooks login', function() {
 
+    const loginPage = new LoginPage()
+    const timeLoggingPage = new TimeLoggingPage()
+
     it('Should display validation for empty user after attempted loggin', function () {
```
linukkis: loggin -> login ;)

PR #64, opened 2019-07-25
Fixed some errors, added more sections describing testing stages.

SauliusLipkevicius:
Fixed errors

### DIFF
--- a/cypress/integration/Pagalbinis.js
+++ b/cypress/integration/Pagalbinis.js
@@ -1,0 +1,47 @@
+ // Dictionary with roles and values - number of tabs it has
+ var roles = { "User":1, "Team Lead":2, "Manager":5, "Accountant":5, "Admin":6};
+  
+ //Mistake in the name
+ var name = "Saulius Lipkevičius"; 
+ var trueName = "Saulius Lipkevicius";
+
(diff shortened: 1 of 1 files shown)


cypress/integration/Pagalbinis.js
```diff
@@ -0,0 +1,47 @@
...
+ var roles = { "User":1, "Team Lead":2, "Manager":5, "Accountant":5, "Admin":6};
+  
+ //Mistake in the name
+ var name = "Saulius Lipkevičius"; 
+ var trueName = "Saulius Lipkevicius";
+
+ it('Should display validation for empty user after attempted loggin', function () {
+            
+    cy.visit('/');
+    cy.get('.Select.not-valid').should('not.visible')
```
linukkis: Many parts of your code still do not use the page object model, for example here: if the validation error class changes in the future for some reason, e.g. into ".Select.not-valid-error", you'd have to go through all the places where you'd used it and update it one by one, whereas if you extracted this into a separate method to retrieve it, you'd only have to change the class name in the method once, thus reducing the need for maintenance greatly :)

This applies to other parts as well, such as cy.get('[id="loginForm.role"]'), cy.get('[type="submit"]'), cy.get('.user-info__title') and others

PR #67, opened 2019-07-25
TC-9 Page Object Model compliant

egle-buce:


cypress/integration/login.js
```diff
@@ -1,105 +1,111 @@
+import LoginPage from '../objects/loginPage'
+import TimeLog from '../objects/timeLog'
+
+const loginPage = new LoginPage()
+const timeLog = new TimeLog()
```
Imlus: change this to timeLogPage

cypress/objects/timeLog.js
```diff
@@ -0,0 +1,37 @@
+class TimeLog {
+
+    getUrl(){
+        return cy.url()
+    }
+    
+
```
Imlus: whya re there double empty rows here?

PR #69, opened 2019-07-25
Laurynas kochas/feature/t 9 refactor t use page object model

lawrynuxaz:


cypress/integration/login.js
```diff
@@ -1,26 +1,32 @@
+import LoginPage from '../objects/loginPage';
+import TimeLogging from '../objects/timeLogging/timelogging'
+const loginPage = new LoginPage()
+
+
```
linukkis: Why so much space here? I think one line would be enough

cypress/integration/login.js
```diff
@@ -1,28 +1,111 @@
...
+        cy.url().should('include', '/time-logging');
+        cy.get('.page__title').contains('Timesheets');
+        cy.get('.calendar').should('be.visible');
+        cy.get('.calendar--today').contains(new Date().getDate());
+        cy.get('.tile.form').should('be.visible');
+        cy.get('.user-info__title').contains('Laurynas Kochas');
+        cy.get('.main-nav').find('li').should('have.length', 2);
+        cy.get('.main-nav__link--active').should('have.css', 'color', 'rgb(64, 76, 237)');
+
+})
```
linukkis: Indent issues starting from here and going down. Increase the indent to the right by one tab.

cypress/integration/login.js
```diff
@@ -1,28 +1,111 @@
...
+        loginPage.visit()
+        loginPage.getUser().click({force:true});
+        loginPage.getName('Laurynas Kochas').click();
+        loginPage.getRole().click({force:true});
+        loginPage.getRoleLabel('User').click();
+        loginPage.getLoginButton().click();
+
+        cy.url().should('include', '/time-logging');
+        cy.get('.page__title').contains('Timesheets');
+        cy.get('.calendar').should('be.visible');
```
linukkis: So everything's great, you've created some nice functions to retrieve the required elements from the Time Logging page, but then forgot to actually use them in your test, right? :)

PR #71, opened 2019-07-25
Ugne lekstutiene/feature/t 9 refactor t use page object model

Cheekiuz:


cypress/objects/timeLoginPage.js
```diff
@@ -0,0 +1,38 @@
...
+
+    getCalendarVisable() {
+        return cy.get('.calendar').should('be.visible')
+    }
+
+    getCalendarToday() {
+        return cy.get('.calendar--today').contains(new Date().getDate())
+    }
+
+    getFileForm() {
```
linukkis: This function's name should be way more descriptive, something like "`getTimeEntryForm()`" (because that's what it is, right?). This way you'll know for sure what this function is meant to get :)

cypress/integration/login.js
```diff
@@ -1,27 +1,106 @@
...
+        loginPage.getUserForm().click({force:true})
+        loginPage.getUserOption("Ugne Lekstutiene").click()
+        loginPage.getLoginForm().click({force:true})
+        loginPage.getLabel('User').click()
+        loginPage.getSubmit().click()
+        timeLoginPage.getURL('.time-logging')
+        timeLoginPage.getPageTitle('Timesheets')
+        timeLoginPage.getCalendarVisable('be.visible')
+        timeLoginPage.getCalendarToday(new Date().getDate())
+        timeLoginPage.getFileForm('be.visible')
```
linukkis: Here and everywhere else where you're using this function the parameter 'be.visible' is not needed, as the function does not even take any parameters. Better leave the brackets empty.

cypress/objects/timeLoginPage.js
```diff
@@ -0,0 +1,38 @@
+class TimeLoginPage {
+    visit() {
+        cy.visit('/')
+    }
+
+    getURL() {
+        return cy.url().should('include', '/time-logging')
```
linukkis: Right now this function not only gets the URL address, but it also asserts that it contains 'time-logging'. I'd probably drop the "should" part and add it in the test cases instead of this function's definition.

cypress/objects/timeLoginPage.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    getPageTitle() {
```
linukkis: According to the name of the function, the function should only get the element and not assert that it contains 'Timesheets', so I'd drop the 'contains' part and add it in the test cases instead of this function's definition :)

cypress/objects/timeLoginPage.js
```diff
@@ -0,0 +1,38 @@
...
+        return cy.get('.page__title').contains('Timesheets')
+    }
+
+    getCalendarVisable() {
+        return cy.get('.calendar').should('be.visible')
+    }
+
+    getCalendarToday() {
```
linukkis: Again. you should only get today's date and not immediately assert it, since this is a "get" function, not an "assert" function, right? :) Let's drop the 'contains' part and add it in the test cases instead

cypress/objects/timeLoginPage.js
linukkis: This comment was not addressed

cypress/objects/timeLoginPage.js
```diff
@@ -0,0 +1,38 @@
...
+        return cy.get('.calendar--today')
+    }
+
+    getTimeEntryForm() {
+        return cy.get('.tile.form')
+    }
+
+    getUserInfoTitle() {
+        return cy.get('.user-info__title').contains('Ugne Lekstutiene')
```
linukkis: I'd drop the 'contains' part and add it in the test cases instead of this function's definition :)

cypress/objects/timeLoginPage.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    getMainLength(tabNumber) {
+        return cy.get('.main-nav').find('li').should('have.length', tabNumber)
```
linukkis: I'd drop the 'should' part and add it in the test cases instead

cypress/objects/timeLoginPage.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    getTimeLoggingColored() {
```
linukkis: Not sure if this is the correct way to do it, it would probably be better to get not the 'Time Logging' tab, but the active tab. Or if you want to only consider the 'Time Logging' tab, then maybe rename this function to add the color assertion, e.g. `assertTimeLoggingColoredBlue()`

cypress/integration/login.js
```diff
@@ -1,27 +1,103 @@
...
+        loginPage.getLabel('User').click()
+        loginPage.getSubmit().click()
+        timeLoginPage.getURL().should('include', '/time-logging')
+        timeLoginPage.getPageTitle('Timesheets')
+        timeLoginPage.getCalendarVisable().should('be.visible')
+        timeLoginPage.getCalendarToday().contains(new Date().getDate())
+        timeLoginPage.getTimeEntryForm('be.visible')
+        timeLoginPage.getUserInfoTitle('Ugne Lekstutiene')
+        timeLoginPage.getMainLength('1')  
+        timeLoginPage.getTimeLoggingColored('rgb(64, 76, 237)')
```
linukkis: There's no point in inserting the RGB color parameter here since this function does not even take any parameters :) You already assert the blue color in the function definition

PR #72, opened 2019-07-25
code refactoring

JevgenijaM:


cypress/integration/login.js
```diff
@@ -1,28 +1,114 @@
+import LoginPage from '../objects/logingPage'
+import TimeLogingPage from '../objects/timeLogingPage'
+const loginPage = new LoginPage()
+const timeLogingPage = new TimeLogingPage()
 describe('Sourcebooks login', function() {
```
linukkis: Feels like this place could use an empty new line to separate these variables from the describe block

cypress/integration/login.js
```diff
@@ -1,28 +1,114 @@
...
 
     it('Should display validation for empty user after attempted loggin', function () {
         
-        cy.visit('/');
-        cy.get('.Select.not-valid').should('not.visible')
-        cy.get('[type="submit"]').click();
-        cy.get('.Select.not-valid').should('be.visible')
+
+        loginPage.visit();
+
```
linukkis: Why the empty line? :)

cypress/objects/timeLogingPage.js
```diff
@@ -0,0 +1,27 @@
...
+    getCalendarToday(){
+        return cy.get('.calendar--today')
+    }
+    getTileForm(){
+        return cy.get('.tile.form')
+    }
+    getInfoTitle(){
+        return cy.get('.user-info__title')
+    }
+    getLength(){
```
linukkis: This could be more descriptive, for example `getNavigationTabCount()` or something similar, because right now it's not really clear whose length you're getting :)

cypress/integration/login.js
```diff
@@ -1,28 +1,114 @@
...
+        timeLogingPage.getCalendarVisible().should('be.visible')
+        timeLogingPage.getCalendarToday().contains(new Date().getDate());
+        timeLogingPage.getTileForm().should('be.visible')
+        timeLogingPage.getInfoTitle().contains('Jevgenija Malomuž');
+        timeLogingPage.getLength().find('li').should('have.length', 1);
+        timeLogingPage.getTimeLoggingColor().contains('Time Logging').should('have.css' , 'color' , 'rgb(64, 76, 237)');
+
+    })
+
+    it('Should be able to login with role User', function () {
```
linukkis: This test case should say 'role Team Lead' and not 'User', right? And then the ones below should use the corresponding roles in their names too

cypress/objects/logingPage.js
```diff
@@ -0,0 +1,24 @@
+class LoginPage{
+    visit(){
+        cy.visit('/')
+    }
+    getValidationError(){
+        return cy.get('.Select.not-valid')
+    }
+    getSubmit(){
```
linukkis: It would be great to add 'Button' to the end, as in `getSubmitButton()`. This way you'll know for sure what kind of element you're talking about, in this case a button :)

cypress/objects/logingPage.js
```diff
@@ -0,0 +1,24 @@
...
+    getUserForm(){
+        return cy.get('[id="loginForm.userId"]')
+    }
+    getUserOption(userName){
+        return cy.get(`[aria-label="${userName}"]`)
+    }
+    getLoginForm(){
+        return cy.get('[id="loginForm.role"]')
+    }
+    getLabel(role){
```
linukkis: What label is this referring to? Please add some more details to what label/field etc. this is referring.

cypress/integration/login.js
```diff
@@ -1,28 +1,114 @@
...
+const timeLogingPage = new TimeLogingPage()
 describe('Sourcebooks login', function() {
 
     it('Should display validation for empty user after attempted loggin', function () {
         
-        cy.visit('/');
-        cy.get('.Select.not-valid').should('not.visible')
-        cy.get('[type="submit"]').click();
-        cy.get('.Select.not-valid').should('be.visible')
+
```
linukkis: Why the double space?

cypress/objects/timeLogingPage.js
```diff
@@ -0,0 +1,27 @@
...
+        return cy.get('.main-nav')
```
linukkis: You could actually append "`.find('li')`" to this :)

cypress/objects/timeLogingPage.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+    getTimeLoggingColor(){
```
linukkis: This does not retrieve the color of the Time Logging tab, it only returns the active tab, so this should probably be renamed to something like `getActiveTab()`

PR #73, opened 2019-07-25
Olga pikul/feature/final

sattixa:


cypress/integration/login.js
```diff
@@ -19,90 +19,92 @@ describe('Sourcebooks login', function() {
...
-        cy.get('.calendar--today').contains(new Date().getDate());
-        cy.get('.main-nav__link.main-nav__link--active').contains('Time Logging').should ('have.css','color','rgb(64, 76, 237)');
+        timelog.includer();
+        timelog.pagetitle().contains('Timesheets')
+        timelog.calendarv().should('be.visible')
+        timelog.getTile().should('be.visible')
+        timelog.getUserTitle().contains("Olga Pikul")
+        timelog.mainNav().should('have.length', 5);
+        timelog.celendarToday().contains(new Date().getDate());
+        timelog.selectedTimelogging().contains('Time Logging').should('have.css' , 'color' , 'rgb(64, 76, 237)');
```
linukkis: Not critical at all, but here's a double space :)

cypress/objects/timelog.js
```diff
@@ -6,7 +6,23 @@ cy.url().should('include', '/time-logging');
 pagetitle(){
 return cy.get('.page__title')
 }
-
+calendarv(){
```
linukkis: Is this a typo? :)

cypress/objects/timelog.js
```diff
@@ -6,7 +6,23 @@ cy.url().should('include', '/time-logging');
...
+return cy.get('.calendar')
+}
+getTile(){
```
linukkis: What tile is this? If this is the time entry form then please update the name of the function to reflect that :) 

Something like `getTimeEntryForm`

cypress/objects/timelog.js
```diff
@@ -6,7 +6,23 @@ cy.url().should('include', '/time-logging');
...
+    return cy.get('.tile.form');
+}
+getUserTitle(){
+    return cy.get('.user-info__title');
+}
+mainNav(){
```
linukkis: The function name tells us nothing because it's not descriptive. If this function is meant to retrieve all the tabs from the main navigation bar, then rename it to something like "`getNavigationTabs()`"

cypress/objects/timelog.js
```diff
@@ -6,7 +6,23 @@ cy.url().should('include', '/time-logging');
...
+    return cy.get('.main-nav').find('li');
+}
+celendarToday(){
```
linukkis: celendar -> calendar? :)

cypress/objects/timelog.js
```diff
@@ -6,7 +6,23 @@ cy.url().should('include', '/time-logging');
...
+    return cy.get('.calendar--today');
+}
+selectedTimelogging(){
```
linukkis: This does not get the time logging tab, it retrieves the _selected_ tab, so better rename this to `getSelectedTab()` or something :)

cypress/objects/loginPage.js
```diff
@@ -3,7 +3,7 @@ class LoginPage {
       cy.visit('/')
 }
 getvalidationerror(){
-    return cy.get('Select.not-valid');
+return cy.get('Select.not-valid');
 
 }
 submitlogin(){
```
linukkis: I'd add "button" at the end so that we know what kind of element we're dealing with here, a button in this case :)

cypress/objects/loginPage.js
```diff
@@ -3,7 +3,7 @@ class LoginPage {
       cy.visit('/')
 }
 getvalidationerror(){
-    return cy.get('Select.not-valid');
+return cy.get('Select.not-valid');
```
linukkis: Please add back the indentation for correct formatting here and everywhere else analogously

cypress/objects/loginPage.js
```diff
@@ -3,7 +3,7 @@ class LoginPage {
       cy.visit('/')
 }
```
linukkis: Starting from here right down to the second-to-last bracket, all code should be indented to the right by one tab. That would be the correct way to format your code :)

cypress/objects/loginPage.js
```diff
@@ -27,12 +27,25 @@ getRole(){
 return cy.get('[id="loginForm.role"]');
 }
 
-roleLabel(){
+roleLabeluser(){
```
linukkis: Do here what you did with getUserOption(): insert a parameter to this function:

> getUserRoleLabel(roleName){
> 	return cy.get(\`[aria-label="${roleName}"]\`);
> }

And then you won't need the other 4 functions, since this one will be universal :)

cypress/objects/timelog.js
sattixa: Name of function?

linukkis: Yeah, 'calendarv' instead of 'calendar' :) But even then it's probably better to add 'get' at the beginning: `getCalendar()`, so that you can immediately understand what the function does when you see it somewhere later.

cypress/objects/timelog.js
sattixa: Ooops, taip ;)

cypress/objects/timelog.js
sattixa: changed! Trying to make right pull request for review.

PR #74, opened 2019-07-25
Finishing testcase 2.

TheMinikit:


cypress/objects/loginPage.js
```diff
@@ -0,0 +1,15 @@
+class LoginPage {
+    visit() {
+        cy.visit('/')
+    }
+
+    getValidationError() {
```
linukkis: Great job! But why didn't you extract other elements into functions as well? :)

cypress/integration/login.js
```diff
@@ -1,27 +1,42 @@
+import LoginPage from '../objects/loginPage'
+
+
```
linukkis: Why the double gap here?

cypress/integration/login.js
```diff
@@ -1,27 +1,42 @@
...
+        loginPage.getValidationError().should('be.visible')
     })
 
-    it('Should be able to login with role User', function () {
+    //T-2
+    var Roles = ["User", "Team Lead", "Manager", "Accountant", "Admin"];
+    for(let i = 0; i < Roles.length; i++)
+    {
+        it(`Should be able to login with role ${Roles[i]}`, function () {
+            var RoleTabNumber = [1, 2, 5, 5, 6];
```
linukkis: You could extract this somewhere outside the 'for' loop, or better yet, attach those numbers to the roles and turn them into key-value pairs in just one array, because right now you're assigning this array 5 times in total, that is once every iteration

cypress/integration/login.js
```diff
@@ -1,27 +1,42 @@
...
+            const loginPage = new LoginPage()
+            loginPage.visit();
+
+            cy.get('[id="loginForm.userId"]').click({force:true});
```
linukkis: It'd be great if you extracted this whole block into functions according to the page object model :)

PR #93, opened 2019-07-31
Dovile Martinkute/feature/TC-4-admin_creates_new_client

mdovile:


nikolaj-tolkaciov (changes requested):
It is hard to perform code review when there is so many un-merged changes from previous test cases, hard to spot what is related specifically to this test case

PR #94, opened 2019-07-31
admin creates new client

RMNT:


cypress/integration/admin.js
```diff
@@ -26,4 +26,22 @@ describe('Sourcebooks admin', function() {
         check.searchTaskName('[type="text"]', task[0]);
         check.visibilityCheck('[role="row"]');
     })
+
+    it('Admin creates a client', function() {
+        var task = [Math.random().toString(36).substring(2, 15), Math.random().toString(36).substring(2, 15), Math.random().toString(36).substring(2, 15), Math.random().toString(36).substring(2, 15), Math.random().toString(36).substring(2, 15)]
```
nikolaj-tolkaciov, line 31: please consider using key value arrays, in order to not use task[3] but instead task["email"] would be more readable and maintainable in a future

nikolaj-tolkaciov: also format this line to be like :
var task = [
     Math.random().toString(36).substring(2, 15), 
     Math.random().toString(36).substring(2, 15),
     .....
]
much easier to read that way

cypress/integration/admin.js
```diff
@@ -26,4 +26,22 @@ describe('Sourcebooks admin', function() {
...
+        task[3] += "@gmail.com";
+        console.log(task[3]);
+
+        check.click('[href="/clients"]');
+        check.click('.btn');
+        check.inputText('[id="clientDetailsForm.organization"]' ,task[0]);
```
nikolaj-tolkaciov, line 37: locators are defined in a test case, this is not good practice according to page object model

cypress/objects/checks.js
```diff
@@ -38,6 +38,10 @@ class checks {
     searchTaskName(nameTitle, text) {
         cy.get(nameTitle).eq(0).type(text);  
     }
+
+    searchOrganizationName() {
```
nikolaj-tolkaciov, line 42: This function is not used and also is not working because it is missing nameTitle and text parameters

nikolaj-tolkaciov (changes requested):

PR #95, opened 2019-07-31
second try to create a pull request

RMNT:


nikolaj-tolkaciov (changes requested):
cant see any changes or actual test case

PR #114, opened 2019-08-01
added date check

kasteckis:


cypress/integration/login.js
```diff
@@ -8,20 +8,28 @@ describe('Sourcebooks login', function() {
...
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
-        cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.user-info__title').contains('Valentinas Kasteckis');
+        cy.get('.main-nav').find('li').should('have.length', 2);
+    })
+
+    it('Date should match todays date', function() {
+        var d = new Date();
+        cy.get('.calendar--selected').find('span').contains(d.getDate());
```
linukkis: Since there are two "span" tags for a given day, it would be better to get the ".calendar__date" :)

kasteckis: @linukkis But calendar__date is always at the top and find() method returns first selected.

cypress/integration/login.js
```diff
@@ -8,20 +8,29 @@ describe('Sourcebooks login', function() {
...
     })
+
+    it('Date should match todays date', function() {
+        var d = new Date();
+        cy.get('.calendar--selected').find('span').get(".calendar__date").contains(d.getDate());
+    })
+
+  
+
+
```
DavidV44: That a lot of unneeded empty lines :)

cypress/integration/login.js
```diff
@@ -8,20 +8,29 @@ describe('Sourcebooks login', function() {
...
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
-        cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.user-info__title').contains('Valentinas Kasteckis');
+        cy.get('.main-nav').find('li').should('have.length', 2);
     })
+
+    it('Date should match todays date', function() {
+        var d = new Date();
+        cy.get('.calendar--selected').find('span').get(".calendar__date").contains(d.getDate());
```
DavidV44: _.get(".calendar__date")_ is not needed for this to work correctly, _.calendar--selected_ is enough to find correct element

DavidV44 (changes requested):

cypress/integration/login.js
linukkis: Yes, but imagine that changed, e.g. a new "span" tag appeared above the date. I understand what you're saying and sometimes we indeed have no other choice other than select the first element (e.g. when all children nodes have no distinct classes or attributes and are very similar), but if we can distinctively retrieve the element we want, it would be better to do it

kasteckis: @linukkis I did what you suggested, but @DavidV44 said ".get(".calendar__date") is not needed for this to work correctly, .calendar--selected is enough to find correct element".

cypress/integration/login.js
kasteckis: I wouldn't agree, if in the later design stage first element wouldn't be ".calendar__date" then this test case would be incorrect.

cypress/integration/login.js
linukkis: I'll have a word with him :D It's just that if you write `cy.get('.calendar--selected').contains(d.getDate());` and want to see if the selected day is 1, but the selected day is actually 2 with 1 hour logged, I'm almost sure the test would pass even though the date is incorrect, because it would assert the logged hours as well

linukkis (approved):
It's just details what we were discussing, because overall it's a great job 👍

PR #118, opened 2019-08-01
Fnished T1 to Check Todays Date1

rasakap:


cypress/integration/login.js
```diff
@@ -14,14 +14,16 @@ describe('Sourcebooks login', function() {
...
 
         cy.url().should('include', '/time-logging');
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
         cy.get('.user-info__title').contains('Demo User');
-        cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.main-nav').find('li').should('have.length', 2);
+
+        cy.get('.calendar--today').find('.calendar__date').contains('1');
```
DavidV44: This test would not work on most days, you should get today's date dynamically and then assert it :)

DavidV44 (changes requested):

PR #120, opened 2019-08-01
T1 date added

JauniusM:


linukkis (commented):
This pull request should point not to the global develop branch, but to your local develop at `Jaunius_Mykolaitis/develop` :)

cypress/integration/login.js
```diff
@@ -23,5 +23,7 @@ describe('Sourcebooks login', function() {
         cy.get('.tile.form').should('be.visible')
         cy.get('.user-info__title').contains('Demo User');
         cy.get('.main-nav').find('li').should('have.length', 1);
+
+        cy.get('.calendar--today').find('span').contains('1');
```
linukkis, line 27: In your next pull request try to use built-in JavaScript functions to retrieve today's date, because this test will not work tomorrow :)

linukkis: Also, instead of finding "span" I'd suggest you get ".calendar__date", because there are two spans :)

PR #125, opened 2019-08-01
Validate today's date

kamilestu:


cypress/integration/login.js
```diff
@@ -24,4 +24,11 @@ describe('Sourcebooks login', function() {
         cy.get('.user-info__title').contains('Demo User');
         cy.get('.main-nav').find('li').should('have.length', 2);
     })
+    
+    it('Validate today\'s date', function () {
```
linukkis, line 28: This is great! One small thing: if this is meant to be a separate standalone test, I think it'd be best if you added the initial login steps to the beginning of this test case :)

PR #126, opened 2019-08-01
Changed Demo User to Team Lead

ghost:
Added date check

### DIFF
--- a/cypress/integration/login.js
+++ b/cypress/integration/login.js
@@ -26,7 +26,9 @@ describe('Sourcebooks login', function() {
     })
 
     it('Should show calendar date', function () {
-        cy.get('.calendar--today').find('span').contains(1);
+        var d = new Date();
+        cy.get('.calendar--today').find('span').contains(d.getDate());
+
     })
 
 })
--- a/cypress/integration/login.js
+++ b/cypress/integration/login.js
@@ -26,7 +26,7 @@ describe('Sourcebooks login', function() {
     })
 
     it('Should show calendar date', function () {
-        cy.get('.calendar--selected').find('span').contains(1);
+        cy.get('.calendar--today').find('span').contains(1);
     })
 
 })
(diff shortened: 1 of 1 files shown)


cypress/integration/login.js
```diff
@@ -8,20 +8,25 @@ describe('Sourcebooks login', function() {
...
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
-        cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.user-info__title').contains('Ieva Stonkaite');
+        cy.get('.main-nav').find('li').should('have.length', 2);
     })
+
+    it('Should show calendar date', function () {
+        cy.get('.calendar--selected').find('span').contains(1);
```
DavidV44: this check selected day, this would not work if another day would be selected. For maximum swag point i suggest asserting that selected day is today's date :)

DavidV44 (changes requested):

cypress/integration/login.js
```diff
@@ -8,20 +8,25 @@ describe('Sourcebooks login', function() {
...
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
-        cy.get('.main-nav').find('li').should('have.length', 1);
+        cy.get('.user-info__title').contains('Ieva Stonkaite');
+        cy.get('.main-nav').find('li').should('have.length', 2);
     })
+
+    it('Should show calendar date', function () {
+        cy.get('.calendar--today').find('span').contains(1);
```
DavidV44: Also this test would not work tomorrow , instead of "1" you could get today's date and pass it like so:
var d = new Date();
cy.get('.calendar--today').find('span').contains(d.getDate());

DavidV44 (changes requested):

DavidV44 (approved):

PR #129, opened 2019-08-01
T2-verifyed all user roles in a loop

MariusLastauskas:


cypress/integration/login.js
```diff
@@ -8,23 +8,36 @@ describe('Sourcebooks login', function() {
         cy.get('.Select.not-valid').should('be.visible')
     })
 
-    it('Should be able to login with role Team Lead', function () {
+    // Tab count test
```
DavidV44: I would personally delete this comment because its clear that this is a roles array and it is used for roles login test, other than that is all good :)

DavidV44 (approved):

PR #130, opened 2019-08-01
changed user in t1 and finished t2

JelenaCernysova:


cypress/integration/t-2.js
```diff
@@ -0,0 +1,97 @@
...
+        cy.get('[aria-label="Admin"]').click();
+        cy.get('[type="submit"]').click();
+
+        cy.url().should('include', '/time-logging');
+        cy.get('.page__title').contains('Timesheets')
+        cy.get('.calendar').should('be.visible')
+        cy.get('.tile.form').should('be.visible')
+        cy.get('.user-info__title').contains('Jelena Cernyšova')
+        cy.get('.main-nav').find('li').should('have.length', 6)
+        cy.get('a[class="main-nav__link main-nav__link--active"]').contains('Time Logging')
```
linukkis: This should be updated to `cy.get('.main-nav').find('li')......`, so that we can get the whole navigation bar and its items

cypress/integration/t-2.js
```diff
@@ -0,0 +1,97 @@
...
+        cy.get('a[class="main-nav__link main-nav__link--active"]').should('have.css', 'color').and('eq', 'rgb(64, 76, 237)');
```
linukkis, line 19: Same as the line above

cypress/integration/t-2.js
```diff
@@ -0,0 +1,97 @@
...
+        cy.get('[aria-label="Accountant"]').click();
+        cy.get('[type="submit"]').click();
+
+        cy.url().should('include', '/time-logging');
+        cy.get('.page__title').contains('Timesheets')
+        cy.get('.calendar').should('be.visible')
+        cy.get('.tile.form').should('be.visible')
+        cy.get('.user-info__title').contains('Jelena Cernyšova')
+        cy.get('.main-nav').find('li').should('have.length', 5)
+        cy.get('a[class="main-nav__link main-nav__link--active"]').contains('Time Logging')
```
linukkis: Same as the comments above: use `cy.get('.main-nav').find('li')......` and please update the rest below as well :)

cypress/integration/t-2.js
linukkis: Why not change here as well? And of course the identical lines of code below :)

I'm approving this pull-request, but please change these places in the next pull request :)

linukkis (approved):

PR #132, opened 2019-08-01
Dainius gaidamavicius/feature/t2 verify all user roles

ghost:


cypress/integration/login.js
```diff
@@ -8,20 +8,21 @@ describe('Sourcebooks login', function() {
         cy.get('.Select.not-valid').should('be.visible')
     })
 
-    it('Should be able to login with role User', function () {
+    const roles = ['User', 'Team Lead', 'Manager', 'Accountant', 'Admin'];
+    const menu = ['1', '2', '5', '5', '6'];
+    roles.forEach((role, id) => {
```
DavidV44: you also use id in this loop, but it;s not actually used in test. Also this id would always be 0

cypress/integration/login.js
```diff
@@ -8,20 +8,21 @@ describe('Sourcebooks login', function() {
         cy.get('.Select.not-valid').should('be.visible')
     })
 
-    it('Should be able to login with role User', function () {
+    const roles = ['User', 'Team Lead', 'Manager', 'Accountant', 'Admin'];
+    const menu = ['1', '2', '5', '5', '6'];
+    roles.forEach((role, id) => {
+        it('Verify all user roles can log in and should see appropriate tabs', function() {
```
DavidV44: It would be nice that each test would have specific name. You could do that by adding user role and menu items count in each test name

DavidV44 (changes requested):

cypress/integration/login.js
```diff
@@ -7,21 +7,61 @@ describe('Sourcebooks login', function() {
...
+            cy.get('.user-info__title').contains('Dainius Gaidamavicius'); 
+            cy.get('.main-nav').find('li').should('have.length', role.tabs);                      
+            cy.get('.main-nav__link--active').should('have.css','color', 'rgb(64, 76, 237)');
+        })
+    })  
 
-    it('Should be able to login with role User', function () {
-
-        cy.visit('/');
+   /* it('T-3: Admin creates new task', function () {
```
DavidV44: You should not commit commented out code, if it's commented out probably that means that it should be deleted

cypress/integration/login.js
```diff
@@ -7,21 +7,61 @@ describe('Sourcebooks login', function() {
...
+        {
+            name: "Accountant",
+            tabs: '5'
+        },
+        {
+            name: "Admin",
+            tabs: '6'
+        }
+    ];
+
```
DavidV44: unnecessary empty line

cypress/integration/login.js
```diff
@@ -7,21 +7,61 @@ describe('Sourcebooks login', function() {
...
-        cy.get('[aria-label="User"]').click();
+        cy.get('[aria-label="Admin"]').click()
         cy.get('[type="submit"]').click();
+        cy.get('[type="button"]').click();
+        cy.get('[id = "taskDetailsForm.name"]').type('TASK1');
+        cy.get('[id = "taskDetailsForm.description"]').type('TASK11');
+        cy.get('[id = "react-select-8--value-item]').select('Yes')
+    }) */
+
+
```
DavidV44: Many unneeded empty lines

DavidV44 (changes requested):

linukkis (approved):
All good! 👍

PR #133, opened 2019-08-01
t2 verify all user roles

EdvinasKveselys:


cypress/integration/login.js
```diff
@@ -8,20 +8,87 @@ describe('Sourcebooks login', function() {
...
         cy.get('.page__title').contains('Timesheets')
         cy.get('.calendar').should('be.visible')
         cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Edvinas Kveselys');
+        cy.get('.main-nav').find('li').should('have.length', 6);
+
+        const today = new Date();
+        const day = today.getDate();
+        cy.get('.calendar--today').contains(day);
```
linukkis: It would be beneficial to append this to get the ".calendar__date" class as well, because if we leave it like that, the test will assert not only the date, but also the hours logged on that day and pass falsely

cypress/integration/login.js
```diff
@@ -8,20 +8,87 @@ describe('Sourcebooks login', function() {
...
         cy.visit('/');
         cy.get('[id="loginForm.userId"]').click({force:true});
-        cy.get('[aria-label="Demo User"]').click();
+        cy.get('[aria-label="Edvinas Kveselys"]').click();
         cy.get('[id="loginForm.role"]').click({force:true});
-        cy.get('[aria-label="User"]').click();
+        cy.get('[aria-label="Admin"]').click();
         cy.get('[type="submit"]').click();
 
         cy.url().should('include', '/time-logging');
```
linukkis, line 20: I don't think we nee this block, because it seems like we're already asserting we're Admins later on

cypress/integration/login.js
```diff
@@ -8,20 +8,87 @@ describe('Sourcebooks login', function() {
...
+        cy.get('[id="loginForm.userId"]').click({force:true});
+        cy.get('[aria-label="Edvinas Kveselys"]').click();
+        cy.get('[id="loginForm.role"]').click({force:true});
+        cy.get('[aria-label="Manager"]').click();
+        cy.get('[type="submit"]').click();
+        cy.get('.user-info__title').contains('Edvinas Kveselys');
+        cy.get('.main-nav').find('li').should('have.length', 5);
+        cy.get('.main-nav__link--active').should('have.css', 'color').and('eq', 'rgb(64, 76, 237)');
+        cy.get('.main-nav__link--active').should('have.attr', 'href', '/time-logging');
+    })
```
DavidV44, line 67: different "it" block should be separated with empty line

cypress/integration/login.js
```diff
@@ -8,20 +8,87 @@ describe('Sourcebooks login', function() {
...
-        cy.get('.user-info__title').contains('Demo User');
+        cy.get('.user-info__title').contains('Edvinas Kveselys');
+        cy.get('.main-nav').find('li').should('have.length', 6);
+
+        const today = new Date();
+        const day = today.getDate();
+        cy.get('.calendar__date').contains(day);
+    })
+
+    it('Verify "User" can log in and should see appropriate tabs', function() {
```
DavidV44: Name "can see appropriate tabs" imply that you check that he sees specific tabs, but you actually   only check the count

DavidV44 (changes requested):

linukkis (approved):
Great job! 👍

PR #134, opened 2019-08-01
Verify all user roles can log in and should see appropriate tabs

dominykasposkus:


cypress/integration/login.js
```diff
@@ -21,11 +21,50 @@ describe('Sourcebooks login', function() {
...
-    
+
     it('Calendar should show todays date', function(){
         cy.get('.calendar--today').find('.calendar__date').contains((new Date()).getDate());
     })
+
+    it('Verify all user roles can log in and should see appropriate tabs', function(){
+        let Roles = [
+            {
+                "name":"User",
```
DavidV44: You should add a space after ":" so its easier to read actual values because now it all looks stuck together like a long spaghetti .

cypress/integration/login.js
```diff
@@ -21,11 +21,50 @@ describe('Sourcebooks login', function() {
...
+        cy.get('.user-info__title').contains('Dominykas Poškus');
         cy.get('.main-nav').find('li').should('have.length', 2);
     })
-    
+
     it('Calendar should show todays date', function(){
         cy.get('.calendar--today').find('.calendar__date').contains((new Date()).getDate());
     })
+
+    it('Verify all user roles can log in and should see appropriate tabs', function(){
```
DavidV44: It would be very nice to have dynamic name of test, like "Verify that Admin role can login and see 6 tabs" and so on.
You can do that by putting roles object before the test and then writing test name:
_"Verify that" + Roles[i].name + "..."_

DavidV44 (changes requested):

cypress/integration/login.js
nikolaj-tolkaciov: This code review remark is not fixed

nikolaj-tolkaciov (changes requested):

DavidV44 (approved):

PR #135, opened 2019-08-01
Aiste laugalyte/feature/t2 verify all user rolls

aistyska:


cypress/integration/T2_verify_all_user_roles.js
```diff
@@ -0,0 +1,92 @@
+describe('Sourcebooks login', function() {
+
+    it('Should be able to log in with a role admin and check if "Time Logging" color is blue', function() {
```
DavidV44, line 3: All these tests can be written with a lot less code. You could create an object that has all the role name and tabs count and then loop the test several times and pass different role name and tabs counts each time

DavidV44 (changes requested):

cypress/integration/T2_verify_all_user_roles.js
DavidV44: UPDATE:  the way you wrote the tests now are okay, my suggestion is optional, if you want you can merge :)

DavidV44 (approved):

PR #136, opened 2019-08-01
Rasa kapacinskaite/feature/t2 verify user roles

rasakap:


cypress/integration/login.js
```diff
@@ -1,29 +1,121 @@
...
-        cy.get('.page__title').contains('Timesheets')
-        cy.get('.calendar').should('be.visible')
-        cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Demo User');
-        cy.get('.main-nav').find('li').should('have.length', 2);
-        let today = new Date();
-        cy.get('.calendar--today').find('.calendar__date').contains(today.getDate());
-    })
-})
+describe('Sourcebooks login', function() {
```
DavidV44, line 1: You only need one "describe" block that would have many "it" blocks.
You should move all "it" blocks to one "describe"

DavidV44 (changes requested):

PR #138, opened 2019-08-01
T-2: Verify if all user roles can log in and see appropriate tabs

mindaugasmc:


cypress/integration/login.js
```diff
@@ -34,7 +33,108 @@ describe('Sourcebooks login', function() {
     cy.get('[type="submit"]').click();
 
     cy.visit('/time-logging');
-    let date = new Date();
-    cy.get('.calendar--today.calendar--selected').contains(date.getDate());
+    cy.get('.calendar--today.calendar--selected').contains(
+      new Date().getDate()
```
linukkis, line 41: Could've kept it a single line I guess, but it's of no difference :)

linukkis (approved):
You could try to put all those "login and check for tabs" test cases into a loop, but other than that - great job! :)

DavidV44 (approved):
nice

PR #142, opened 2019-08-01
Saule raudyte/feature/t2 verify all user roles

saulerau:


.vscode/tasks.json
```diff
@@ -0,0 +1,12 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
```
linukkis: Not sure what this file is, but I'm pretty sure this is not to be commited :)

linukkis (commented):
Generally speaking all your tests are fine, but it is very inconvenient to put them into different files when they are all doing pretty much the same thing. The best option would probably be to put them back to some single file, could be `login.js`, and create a loop to iterate through all five test cases :)

DavidV44 (changes requested):
login.js should not be deleted as it contains T-1 test

PR #144, opened 2019-08-01
Jaunius mykolaitis/feature/t2 verify all user roles

JauniusM:


DavidV44 (approved):

linukkis (approved):
You could try running a loop to iterate through the five test cases, but other than that - great job! 👍

PR #147, opened 2019-08-01
Agne gadeikiene/feature/t2 verify all user

januagne:


cypress/integration/login.js
```diff
@@ -17,6 +17,8 @@ describe('Sourcebooks login', function() {
         cy.get('[aria-label="User"]').click();
         cy.get('[type="submit"]').click();
 
+
```
linukkis, line 20: This is not a showstopper, but please try to remove 2 of the 3 lines with the next pull-request :)

linukkis (commented):
It is quite inconvenient to separate these 5 test cases into different files, I think what you could do is put them back into a single file like `login.js` and run a loop that iterates through them :) or if it's too difficult, just put them all in a single 'describe' block

PR #151, opened 2019-08-01
i did T3 fully

kasteckis:


cypress/integration/t3.js
```diff
@@ -0,0 +1,23 @@
+describe('Sourcebooks login', function() {
+
+    it('Admin can create new Task', function(){
+        cy.visit('/');
+        cy.get('[id="loginForm.userId"]').click({force:true});
+        cy.get('[aria-label="Valentinas Kasteckis"]').click();
+        cy.get('[id="loginForm.role"]').click({force:true});
+        cy.get('[aria-label="Admin"]').click();
+        cy.get('[type="submit"]').click();
+        cy.visit('/tasks');
```
nikolaj-tolkaciov, line 13: It would be great to assert that task you are about to create is not in the list, and only after creating it it appears there

nikolaj-tolkaciov (changes requested):

PR #154, opened 2019-08-01
T4 test for new client creation added, refactored log in testing and …

MariusLastauskas:
…makeid()

### DIFF
--- a/cypress/integration/login.js
+++ b/cypress/integration/login.js
@@ -1,102 +1,35 @@
 import LoginPage from '../obj/loginPage';
-import ClientsPage from '../obj/clients/clientsPage';
-import ClientCreatePage from '../obj/clients/clientCreatePage';
-import TasksPage from '../obj/tasks/tasksPage';
-import TaskCreatePage from '../obj/tasks/taskCreatePage';
-import TaskEditPage from '../obj/tasks/taskEditPage';
 import TimeLoggingPage from '../obj/timeLogging/timeLoggingPage';
 import Common from '../obj/common';
 
 const loginPage = new LoginPage();
-const clientsPage = new ClientsPage();
-const clientCreatePage = new ClientCreatePage();
-const tasksPage = new TasksPage();
-const taskCreatePage = new TaskCreatePage();
-const taskEditPage = new TaskEditPage();
 const timeLoggingPage = new TimeLoggingPage();
 const common = new Common();
     
-var roles = [
+const roles = [
     { name: 'User', 'tabCount': 1 }, 
-    { name: 'Team Lead', tabCount: 2 }, 
-    { name: 'Manager', tabCount: 5 }, 
-    { name: 'Accountant', tabCount: 5}, 
-    { name: 'Admin', tabCount: 6 }
+    { name: 'Team Lead', 'tabCount': 2 }, 
+    { name: 'Manager', 'tabCount': 5 }, 
+    { name: 'Accountant', 'tabCount': 5}, 
+    { name: 'Admin', 'tabCount': 6 }
 ];
-
-function makeId() {
-    let timeStamp = new Date().getTime();
-    return 'Random_Name_' + timeStamp;
-}
-
-function testLogIn(role) {
-    loginPage.visit();
-    loginPage.getUserDropdown().click({force:true});
-    loginPage.getUserFromDropdown('Marius Lastauskas').click();
-    loginPage.getRoleDropdown().click({force:true});
-    loginPage.getRoleFromDropdown(role).click();
-    loginPage.getSubmitButton().click();
-}
-
-describe('Sourcebook test Clients', function() {
-    it('Should be available for Admin to create new client', function () {
-        cy.loginAs("Admin");
-
-        const organization = makeId();
-        const email = makeId() + '@mail.com';
-        clientsPage.visit();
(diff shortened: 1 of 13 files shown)


cypress/integration/login.js
```diff
@@ -17,22 +9,45 @@ describe('Sourcebooks login', function() {
...
-    it('Should be able to create new task as Admin', function () {
+    it('Should display validation for empty user after attempted loggin', function () {
+        
+        cy.visit('/');
+        cy.get('.Select.not-valid').should('not.visible')
+        cy.get('[type="submit"]').click();
+        cy.get('.Select.not-valid').should('be.visible')
+    })
+
+    function testLogIn(role) {
```
nikolaj-tolkaciov: It would be best to keep functions separated from test cases or at least not mix them inside describe

cypress/integration/login.js
```diff
@@ -17,22 +9,45 @@ describe('Sourcebooks login', function() {
...
 
     function makeid(length) {
-        var result = '';
-        var characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
-        var charactersLength = characters.length;
-        for (var i = 0; i < length; i++) {
-           result += characters.charAt(Math.floor(Math.random() * charactersLength));
-        }
-        return result;
+        return 'Random_Name_' + Math.floor(Math.random() * 1000000000000);
```
nikolaj-tolkaciov: why you are not using timestamp? It would make it more predictable and event traceable, with random there is still chance...

cypress/integration/login.js
```diff
@@ -17,22 +9,45 @@ describe('Sourcebooks login', function() {
...
         cy.visit('/');
         cy.get('[id="loginForm.userId"]').click({force:true});
         cy.get('[aria-label="Marius Lastauskas"]').click();
         cy.get('[id="loginForm.role"]').click({force:true});
-        cy.get('[aria-label="Admin"]').click();
+        cy.get('[aria-label="' + role + '"]').click();
+        cy.get('[type="submit"]').click();
+    }
+
+    it('Should be available for Admin to create new client', function () {
```
nikolaj-tolkaciov: This should probably go to another file, as it is not really testing "login"

cypress/integration/login.js
```diff
@@ -1,13 +1,5 @@
...
-
-    it('Should display validation for empty user after attempted loggin', function () {
-        
-        cy.visit('/');
-        cy.get('.Select.not-valid').should('not.visible')
-        cy.get('[type="submit"]').click();
-        cy.get('.Select.not-valid').should('be.visible')
-    })
-
+describe('Sourcebooks testLogIn', function() {
```
nikolaj-tolkaciov: From describe block's name I would assume it is testing your created function

cypress/integration/login.js
```diff
@@ -17,22 +9,45 @@ describe('Sourcebooks login', function() {
     ];
 
     function makeid(length) {
```
nikolaj-tolkaciov: According to coding convention we should use camel case, so it should be "makeId"

nikolaj-tolkaciov: Also length parameter is not used

cypress/integration/login.js
```diff
@@ -17,22 +9,45 @@ describe('Sourcebooks login', function() {
...
+        cy.get('[id="clientDetailsForm.organization"]').type(organization);
+        cy.get('[id="clientDetailsForm.contacts_firstName_0"]').type(makeid(20));
+        cy.get('[id="clientDetailsForm.contacts_lastName_0"]').type(makeid(20));
+        cy.get('[id="clientDetailsForm.contacts_email_0"]').type(makeid(20) + '@mail.com');
         cy.get('[type="submit"]').click();
+        
+        cy.visit('/clients');
+        cy.get('[class="field--filter"]').first().find('input').type(organization);
+        cy.get('[class="ag-body-viewport-wrapper"').find('.ag-body-viewport').should('have.length', 1);
+    })
```
nikolaj-tolkaciov: Overall your client is consisting of bunch of numbers, it would be wise to at least prefix firstName+makeid() and so on.
Also aserting for created values would be nice too in order to confirm customer is created with expected values

nikolaj-tolkaciov (changes requested):

MariusLastauskas:
Fixed comments while refactoring T9

cypress/integration/tasks.js
```diff
@@ -0,0 +1,32 @@
...
+        taskCreatePage.getBillableDropdown().click();
+        taskCreatePage.getBillableFromDropdown('Yes').click();
+        taskCreatePage.getFormRate().clear();
+        taskCreatePage.getFormRate().type(Math.random() * 10);
+        taskCreatePage.getSubmitButton().click();
+
+        taskEditPage.checkUrl();
+        tasksPage.visit();
+        tasksPage.getNameFilter().type(taskName);
+        tasksPage.getFilteredList().should('have.length', 1);
```
DavidV44, line 31: You should also check that task that you just created have same name description and so on, because there is possibility that somehow task with wrong description or something could be created and the test would still pass

DavidV44 (changes requested):

PR #157, opened 2019-08-01
Saule raudyte/feature/t3 admin creates new task

saulerau:
admin creates task


cypress/integration/create.js
```diff
@@ -0,0 +1,29 @@
+describe('Sourcebooks login', function() {
+
+    it('Admin should create a new task', function () {
+
+        const taskName = Math.ceil((Math.random())*10000);
```
DavidV44: Making a random name will work 99.99% of time for not getting a duplicate name, even better approach would be something like:
`const taskName = 'task ' + Date.now();`
`Date.now()` gives us a number of milliseconds from January 1, 1970 which is better than a random number (random number could theoretically repeat). This is a minor issue :)

cypress/integration/create.js
```diff
@@ -0,0 +1,29 @@
...
+        cy.get('[aria-selected="true"]').click({ force: true }).get('[aria-label="Yes"]').click({ force: true });
+        cy.get('[id="taskDetailsForm.rate"]').clear().type("5.000");
+        cy.get('[type="submit"]').click({force:true});
+
+        cy.wait(1000);
+        cy.url().should('to.not.equal',('/tasks/create'));
+        
+        cy.visit('/tasks');
+        cy.get('.field__text').first().click().type(taskName);
+        cy.get('[title=' + taskName + ']').should('exist');
```
DavidV44: It's always better to check if all fields are as expected, because checking the name does not guarantee that task description is as expected. For now this is okay, but this is something to keep in mind for future :)

DavidV44 (changes requested):
You shouldn't delete your login .js file as it contains T-1 test cases

PR #158, opened 2019-08-01
Saule raudyte/feature/t4 admin creates new client

saulerau:


cypress/integration/create.js
```diff
@@ -0,0 +1,29 @@
+describe('Sourcebooks login', function() {
+
+    it('Admin should create a new client', function () {
+
+        const organizationName = Math.ceil(Math.random()*100000);
```
DavidV44: Making a random name will work 99.99% of time for not getting a duplicate name, even better approach would be something like:
`const taskName = 'task ' + Date.now();`
`Date.now()` gives us a number of milliseconds from January 1, 1970 which is better than a random number (random number could theoretically repeat). This is a minor issue :)

cypress/integration/create.js
```diff
@@ -0,0 +1,29 @@
...
+        cy.get('[aria-label="Saulė Raudytė"]').click();
+        cy.get('[id="loginForm.role"]').click({force:true});
+        cy.get('[aria-label="Admin"]').click();
+        cy.get('[type="submit"]').click();
+
+        cy.get('[href="/clients"]').click();
+        cy.get('.btn').click();
+
+        cy.get('[id="clientDetailsForm.organization"]').type(organizationName);
+        cy.get('[id="clientDetailsForm.contacts_firstName_0"]').type(organizationName+1);
```
DavidV44: It's always better to leave spaces before and after math operators, it improves code readability :)

cypress/integration/create.js
```diff
@@ -0,0 +1,29 @@
...
+        cy.get('[id="clientDetailsForm.contacts_lastName_0"]').type(organizationName+2);
+        cy.get('[id="clientDetailsForm.contacts_email_0"]').type(organizationName+'@email.com');
+        cy.get('[type="submit"]').click();
+        cy.wait(1000);
```
DavidV44, line 22: this wait does not guarantee that test will always pass, what if for test to pass we needed to wait for 1.5s? I would suggest that after submit you check if success pop up was displayed at the top of page and then proceed.

DavidV44 (changes requested):
You should not delete your login.js file as it contains T-1 test cases

PR #161, opened 2019-08-02
refactored code

kasteckis:


cypress/obj/loginPage.js
```diff
@@ -0,0 +1,88 @@
...
+    {
+        return cy.get('[type="submit"]');
+    }
+
+    getSpecificUserFromDropDown(userName)
+    {
+        return cy.get('[aria-label="' + userName + '"]');
+    }
+
+    openUpUserDropDown()
```
linukkis: This function would probably be better called "get" instead "open", because you do not actually click it here in the definition, or you could add "click()" at the end :)

cypress/obj/loginPage.js
```diff
@@ -0,0 +1,88 @@
...
+    {
+        return cy.get('[id="loginForm.userId"]');
+    }
+
+    openUpRoleDropDown()
```
linukkis: Same as above

cypress/obj/loginPage.js
```diff
@@ -0,0 +1,88 @@
...
+    {
+        return cy.get('.page__title');
+    }
+    
+    getCalendar()
+    {
+        return cy.get('.calendar');
+    }
+
+    getTitleForm()
```
linukkis: It's "tile", not "title" ;)

cypress/obj/loginPage.js
```diff
@@ -0,0 +1,88 @@
...
+    {
+        return cy.get('.tile.form');
+    }
+
+    checkIfInfoTitleIsSpecific(nameSurname)
```
linukkis: The name of this function could be more descriptive, e.g. `checkLoginName(nameSurname)`, so that we know from the function name that were talking about checking our name. Checking the "info title" might not seem very clear :)

linukkis (commented):
Very good job, a couple of small fixes and you're good to go :)

cypress/integration/login.js
```diff
@@ -1,122 +1,95 @@
...
-
-        cy.url().should('include', '/time-logging');
-        cy.get('.page__title').contains('Timesheets')
-        cy.get('.calendar').should('be.visible')
-        cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Valentinas Kasteckis');
-        cy.get('.main-nav').find('li').should('have.length', 2);
+        loginPage.visit();
+        loginPage.getUserDropDown().click({force:true});
+        loginPage.getSpecificUserFromDropDown("Valentinas Kasteckis").click();
```
nikolaj-tolkaciov, line 18: As you often use same User it would be wise to extract it as a variable and then reuse it in all instances of this function and loginPage.checkLoginName("Valentinas Kasteckis");

nikolaj-tolkaciov (approved):
Good job! one small suggested improvement, but you can merge it without it

PR #163, opened 2019-08-02
T1 login refactoring

rasakap:


DavidV44 (changes requested):
There is no` loginPage.js` file

cypress/objects/LoginPage.js
```diff
@@ -0,0 +1,63 @@
...
+
+    getUserRoleForm() {
+        return cy.get('[id="loginForm.role"]')
+    }
+
+    getUserRoleFromDropDown(userRole) {
+        return cy.get(`[aria-label="${userRole}"]`)
+    }
+
+    url() {
```
DavidV44: `url` does not tell me what this method does, i suggest naming it `getUrl`

cypress/objects/LoginPage.js
```diff
@@ -0,0 +1,63 @@
...
+
+    getItemCalendar() {
+        return cy.get('.calendar')
+    }
+
+    getFormTitle() {
+        return cy.get('.tile.form')
+    }
+
+    getUserInfoTitle(userName) {
```
DavidV44: Methods like these could be put in `common.js` or `pageComponents.js` files, because this is shared between pages and if you would want to check username or navigation bar it would look weird if you have to use `LoginPage `to do so.

DavidV44 (changes requested):

cypress/objects/LoginPage.js
```diff
@@ -0,0 +1,59 @@
...
+
+    getItemCalendar() {
+        return cy.get('.calendar')
+    }
+
+    getFormTitle() {
+        return cy.get('.tile.form')
+    }
+
+        getMainNavigationBar() {
```
DavidV44: This also should be in common

cypress/objects/LoginPage.js
```diff
@@ -0,0 +1,59 @@
...
+
+    getUserRoleFromDropDown(userRole) {
+        return cy.get(`[aria-label="${userRole}"]`)
+    }
+
+    geturl() {
+        return cy.url()
+    }
+
+    getPageTitle() {
```
DavidV44: This also should be in common

cypress/objects/LoginPage.js
```diff
@@ -0,0 +1,59 @@
...
+
+    getUserRoleForm() {
+        return cy.get('[id="loginForm.role"]')
+    }
+
+    getUserRoleFromDropDown(userRole) {
+        return cy.get(`[aria-label="${userRole}"]`)
+    }
+
+    geturl() {
```
DavidV44: This also should be in common

DavidV44 (changes requested):

DavidV44 (approved):
pretty cool 😎

PR #164, opened 2019-08-02
T9 refactoring in Page Object modal manner

ModestasDaunys:


cypress/objects/loginPage.js
```diff
@@ -0,0 +1,27 @@
+class LoginPage {
+    visit() {
+        cy.visit('/');
+    }
+
+    getUserFieldSelectorById() {
```
DavidV44: Method name could be clearer, because now this method does not return a selector, it returns element which you can click later. I would name this `getUserField`

cypress/objects/loginPage.js
```diff
@@ -0,0 +1,27 @@
...
+        return cy.get('[id="loginForm.userId"]');
+    }
+
+    getRoleFieldSelectorById(){
```
DavidV44: Method name could be clearer, because now this method does not return a selector, it returns element which you can click later. I would name this `getRoleField`

cypress/objects/loginPage.js
```diff
@@ -0,0 +1,27 @@
...
+
+    getUserOptionFromDropdown(user) {
+        return cy.get(`[aria-label="${user}"]`);
+    }
+
+    getRolesOptionFromDropdown(role) {
+        return cy.get(`[aria-label="${role}"]`);
+    }
+
+    getButtonByType(type) {
```
DavidV44: This method should not be in loginPage objects as this is generic method that could be used in other pages.
You should either name this `getLoginButton` and don't pass type to it
or put this method as a helper method in `helpers.js `file
I recommend 1st option

cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,28 @@
+class TimeLoggingPage {
+
+    assertPageTimeLogginDisplayed() {
+        return cy.url().should('include', '/time-logging');
+    }
+
+    getUserProfileButtonSelector() {
```
DavidV44: Here also method naming could be better, as you return an element not a selector

cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,28 @@
...
+        return cy.get('.user-info__title');
+    }
+
+    getNavListSelector() {
+        return cy.get('.main-nav');
+    }
+
+    getActiveNavMenuOptionSelector() {
```
DavidV44: Here also method naming could be better, as you return an element not a selector

cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,28 @@
...
+
+    assertPageTimeLogginDisplayed() {
+        return cy.url().should('include', '/time-logging');
+    }
+
+    getUserProfileButtonSelector() {
+        return cy.get('.user-info__title');
+    }
+
+    getNavListSelector() {
```
DavidV44: Here also method naming could be better, as you return an element not a selector, also navigation list is not tied to specific page, so it would not be logical to use "timeLoggingPage" if you want to check navigation list while in task page. I suggest putting navigation list methods to different file like `pageComponents.js` where you would put all the stuff that is shared between pages

cypress/integration/login.js
```diff
@@ -1,23 +1,42 @@
-describe('Sourcebooks login', function() {
+import LoginPage from '../objects/loginPage';
+import TimeLoggingPage from '../objects/timeLoggingPage';
+
+const loginPage = new LoginPage();
+const timeLoggingPage = new TimeLoggingPage();
 
+function prepareSUT () {
```
DavidV44, line 9: will this method be reused? Because now i think more simple way would be to put all the variable declarations in `describe `block just before `it`

DavidV44 (changes requested):

cypress/objects/loginPage.js
ModestasDaunys: fixed with commit 'f073c5'

cypress/objects/loginPage.js
ModestasDaunys: fixed with commit 'f073c5'

cypress/objects/loginPage.js
ModestasDaunys: fixed with commit 'f073c5'

cypress/objects/timeLoggingPage.js
ModestasDaunys: fixed with commit 'f073c5'

cypress/objects/timeLoggingPage.js
ModestasDaunys: fixed with commit 'f073c5'

cypress/integration/login.js
ModestasDaunys: I don't know the scope of this workshop, so if we are going to implement more tests, preparation of test suite does make sense, therefore I'd rather choose to have it.

linukkis: I think I second what David said, but if you really think you need it, it does work so I reckon there ain't no harm in it.

linukkis (approved):
Awesome job! 👍

cypress/objects/timeLoggingPage.js
ModestasDaunys: fixed with commit 'f073c5'

PR #165, opened 2019-08-02
Refactored tests to use PageObjects

Modkmi:


cypress/integration/login.js
```diff
@@ -1,28 +1,47 @@
...
-                cy.get('.main-nav__link--active').should('have.css', 'color', 'rgb(64, 76, 237)');
+            ', check current day and active tab color', function () {
+
+                loginPage.visit();
+                loginPage.getUserDropDown().click({ force: true });
+                loginPage.getUserOption(userName).click();
+                loginPage.getRoleDropDown().click({ force: true });
+                loginPage.getRoleOption(roles[i]).click();
+                loginPage.getSubmitButton().click();
+
```
DavidV44, line 35: One empty line is enough for code spacing

cypress/objects/TaskPage.js
```diff
@@ -0,0 +1,3 @@
+export const TaskPage = {
```
DavidV44, line 1: You should not commit code that is not used in any way. I Assume this will be used in future, but when that time comes, the commit this file

cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,29 @@
...
+    getUserInfoTitle() {
+        return cy.get('.user-info__title');
+    }
+    getMainNav() {
+        return cy.get('.main-nav');
+    }
+    getCalendarToday() {
+        return cy.get('.calendar--today');
+    }
+    getActiveMainNav() {
```
DavidV44, line 23: same as above

cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,29 @@
...
+    getCalendarSelector() {
+        return cy.get('.calendar');
+    }
+    getTileForm() {
+        return cy.get('.tile.form');
+    }
+    getUserInfoTitle() {
+        return cy.get('.user-info__title');
+    }
+    getMainNav() {
```
DavidV44, line 17: same as above

cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,29 @@
...
+    getPageTitle() {
+        return cy.get('.page__title');
+    }
+    getCalendarSelector() {
+        return cy.get('.calendar');
+    }
+    getTileForm() {
+        return cy.get('.tile.form');
+    }
+    getUserInfoTitle() {
```
DavidV44, line 14: navigation list is not tied to specific page, so it would not be logical to use "timeLoggingPage" if you want to check navigation list while in task page. I suggest putting navigation list methods to different file like `pageComponents.js `where you would put all the stuff that is shared between pages

cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,29 @@
+class TimeLoggingPage {
+    assertUrl(url) {
```
DavidV44, line 2: what if you would want to assert url of task page? In that case you would need to use timeLogginPage object which would be illogical. I suggest either putting this method in file where such common methods would be used, like `common.js`
or by not passing url to this method

cypress/integration/login.js.rej
```diff
@@ -0,0 +1,10 @@
+diff a/cypress/integration/login.js b/cypress/integration/login.js	(rejected hunks)
```
DavidV44, line 1: This file should not be commited

DavidV44 (changes requested):
all the .vs/ folder files should not be commited

PR #169, opened 2019-08-02
Jaunius mykolaitis/feature/t9 refactor page

JauniusM:


cypress/Objects/loginPage.js
```diff
@@ -0,0 +1,26 @@
...
+    getUserValidationIndicator(){
+        return cy.get('.Select.not-valid')
+    }
+    getSubmitButton(){
+        return cy.get('[type="submit"]')
+    }
+    getSpecificUserFromDropDown(userName){
+        return cy.get(`[aria-label="${userName}"]`)
+    }
+    getUserIDLogin(){
```
DavidV44: this method should be named more clearly, like `getUserDropdown`

cypress/Objects/loginPage.js
```diff
@@ -0,0 +1,26 @@
...
+        return cy.get('[id="loginForm.userId"]')
+    }
+    getLoginFormRole(){
```
DavidV44: this method should be named more clearly, like `getRoleDropdown`

cypress/Objects/TimeLoggin.js
```diff
@@ -0,0 +1,28 @@
...
+    getPageTitle(){
+        return cy.get('.page__title')
+    }
+    getCalendar(){
+        return cy.get('.calendar')
+    }
+    getTimeTile(){
+        return cy.get('.tile.form')
+    }
+    getUserInfoFromTitle(){
```
DavidV44: Methods like these could be put in `common.js` or `pageComponents.js` files, because this is shared between pages and if you would want to check username it would look weird if you have to use `timeLogging `to do so.

DavidV44 (changes requested):

cypress/integration/T2.js
```diff
@@ -0,0 +1,118 @@
...
+        loginPage.getSubmitButton().click();
+
+        timeLoggin.urlIncludeTimeLogging().should('include', '/time-logging');
+        timeLoggin.getPageTitle().contains('Timesheets')
+        timeLoggin.getCalendar().should('be.visible')
+        timeLoggin.getTimeTile().should('be.visible')
+        common.getUserName().contains('Jaunius Mykolaitis');
+        timeLoggin.getMainNavWithSpecificLength().should('have.length', 6);
+
+        timeLoggin.getMainNavActiveLinkColor();
```
nikolaj-tolkaciov: you are getting it but not doing anything with it, if you perform validation inside a function it would be nice to mention that in function name

cypress/integration/T2.js
```diff
@@ -0,0 +1,118 @@
...
+        timeLoggin.getTodaysDay();
```
nikolaj-tolkaciov: you are getting it but not doing anything with it, if you perform validation inside a function it would be nice to mention that in function name

cypress/Objects/TimeLoggin.js
```diff
@@ -0,0 +1,25 @@
...
+        return cy.get('.calendar')
+    }
+    getTimeTile(){
+        return cy.get('.tile.form')
+    }
+    getMainNavWithSpecificLength(){
+        return cy.get('.main-nav').find('li')
+    }
+    getMainNavActiveLinkColor(){
+        return cy.get('.main-nav__link--active').should('have.css', 'color').and('equal', 'rgb(64, 76, 237)')
```
nikolaj-tolkaciov: from how you use test case you don't need to return this , it is enough to just execute this line

cypress/Objects/TimeLoggin.js
```diff
@@ -0,0 +1,25 @@
...
+    }
+    getTodaysDay(){
+        return cy.get('.calendar--today').find('span').contains(new Date().getDate())
```
nikolaj-tolkaciov: from how you use test case you don't need to return this , it is enough to just execute this line

cypress/Objects/TimeLoggin.js
```diff
@@ -0,0 +1,25 @@
+class TimeLoggin{
+    urlIncludeTimeLogging(){
+        return cy.url()
```
nikolaj-tolkaciov: from the name of your function I would assume you actually perform some kind of validation that it includes time logging url.

Overall function that calls another function without any parameters is kind of useless, you could just call cy.url() instead.

what I would suggest to do is this approach:
validateUrl() {
      cy.url().should('include', '/time-logging')
}

cypress/integration/login.js
```diff
@@ -1,27 +1,37 @@
...
 
-    it('Should display validation for empty user after attempted loggin', function () {
-        
-        cy.visit('/');
-        cy.get('.Select.not-valid').should('not.visible')
-        cy.get('[type="submit"]').click();
-        cy.get('.Select.not-valid').should('be.visible')
+    beforeEach(function() {
+        cy.loginAs("Admin")
+        cy.visit('/')
```
nikolaj-tolkaciov: if you bypassing login it is not actually testing it

cypress/integration/login.js
```diff
@@ -1,27 +1,37 @@
...
     })
+    /*it('Should display validation for empty user after attempted loggin', function () {
+        
+        loginPage.visit()
+        loginPage.getUserValidationIndicator().should('not.visible')
+        loginPage.getSubmitButton().click();
+        loginPage.getUserValidationIndicator().should('be.visible')
+    })*/
```
nikolaj-tolkaciov: please do not leave commented out test cases in final version, ether delete it or un-comment it

cypress/integration/login.js
```diff
@@ -1,27 +1,37 @@
...
 
     it('Should be able to login with role User', function () {
```
nikolaj-tolkaciov, line 19: Now it is not actually testing User role, it is admin. This loginAs functionality should be used in test case 3 file to avoid using login each time

cypress/integration/T2.js
```diff
@@ -0,0 +1,118 @@
...
+describe('Sourcebooks login', function() {
+    it('Should display validation for empty user after attempted loggin', function () {
+        
+        loginPage.visit();
+        loginPage.getUserValidationIndicator().should('not.visible')
+        loginPage.getSubmitButton().click();
+        loginPage.getUserValidationIndicator().should('be.visible')
+    })
+
+    it('Should be able to login with user Jaunius_Mykolaitis to role Admin', function () {
```
nikolaj-tolkaciov: why there is a '_' between your name? it is not used like that in actual test case :)

cypress/integration/login.js
nikolaj-tolkaciov: also this test case is covered in T2.js file

nikolaj-tolkaciov (changes requested):

cypress/Objects/TimeLoggin.js
```diff
@@ -0,0 +1,25 @@
+class TimeLoggin{
+    validateUrl() {
+        cy.url().should('include', '/time-logging')
+        }
```
nikolaj-tolkaciov, line 4: Indent for this line is too far right, please use same approach as in the rest of the file

cypress/integration/login.js
nikolaj-tolkaciov: this is a duplicate test case of T2.js
it('Should be able to login with user Jaunius Mykolaitis to role User',

cypress/integration/login.js
```diff
@@ -1,27 +1,35 @@
...
+import TimeLoggin from '../objects/TimeLoggin';
+import Common from '../objects/Common';
+const common = new Common();
+const timeLoggin = new TimeLoggin();
+const loginPage = new LoginPage();
+
+
 describe('Sourcebooks login', function() {
 
     it('Should display validation for empty user after attempted loggin', function () {
```
nikolaj-tolkaciov, line 11: this is a duplicate test case of T2.js
it('Should display validation for empty user after attempted loggin'

nikolaj-tolkaciov (changes requested):
Almost there, previous remakr have not been fixed, login.js have same test cases as t2.js and there is no need to keep duplicate test cases

PR #170, opened 2019-08-02
Code refactored, page classes added

MariusLastauskas:


cypress/obj/timeLogging/timeLoggingPage.js
```diff
@@ -0,0 +1,35 @@
...
+
+    getPageTitle() {
+        return cy.get('.page__title');
+    }
+
+    getCalendar() {
+        return cy.get('.calendar');
+    }
+
+    getFormTitle() {
```
linukkis: It's "tile", not "title" ;)

cypress/obj/tasks/tasksPage.js
```diff
@@ -0,0 +1,19 @@
+class TaskSPage {
```
linukkis: Is there a reason the "S" is capital?

cypress/obj/clients/clientCreatePage.js
```diff
@@ -0,0 +1,23 @@
...
+
+    getFormLastName() {
+        return cy.get('[id="clientDetailsForm.contacts_lastName_0"]');
+    }
+
+    getFormEmail() {
+        return cy.get('[id="clientDetailsForm.contacts_email_0"]');
+    }
+
+    getFormSubmit() {
```
linukkis: I believe this is a button, right? To keep consistent, let's add `...Button()` at the end so that we know it's a button we're interacting with :) I see you've done that in other places already 👍

linukkis (commented):
Everything is great, you really got the hang of this 👍  but why exactly are you merging T9 into T4 and not your 'develop'? Is that on purpose?

cypress/integration/login.js
```diff
@@ -1,92 +1,116 @@
...
-        { name: 'Manager', tabCount: 5 }, 
-        { name: 'Accountant', tabCount: 5}, 
-        { name: 'Admin', tabCount: 6 }
-    ];
+import LoginPage from '../obj/loginPage';
+import ClientsPage from '../obj/clients/clientsPage';
+import ClientCreatePage from '../obj/clients/clientCreatePage';
+import TasksPage from '../obj/tasks/tasksPage';
+import TaskCreatePage from '../obj/tasks/taskCreatePage';
+import TasksEditPage from '../obj/tasks/taskEditPage';
```
DavidV44: I don't think this is used anywhere :)

MariusLastauskas:
I had made changes to T4 branch, but pull request had some comments so I decided to branch from T4 and refactor, fix those bugs, then merge in T4 and then deal with T4 bugs. :)

cypress/obj/timeLogging/timeLoggingPage.js
```diff
@@ -0,0 +1,35 @@
...
+
+    getCalendar() {
+        return cy.get('.calendar');
+    }
+
+    getFormTile() {
+        return cy.get('.tile.form');
+    }
+
+    getUserNameTitle() {
```
DavidV44: Methods like these could be put in` common.js` or` pageComponents.js` files, because this is shared between pages and if you would want to check username it would look weird if you have to use `timeLoggingPage` to do so.

DavidV44 (changes requested):

cypress/integration/login.js
MariusLastauskas: They are used later on :)
in lines 31-38, 77-79, 68-74,  79

DavidV44:
```
import TasksEditPage from '../obj/tasks/taskEditPage';
import TaskEditPage from '../obj/tasks/taskEditPage';
```
you use `TaskEditPage`, but `TasksEditPage` is not used. And even if you would use `TasksEditPage` it would be bad, because you should not use 2 objects that are same

MariusLastauskas: okey, thank you :)

nikolaj-tolkaciov: This is still not used anywhere

cypress/integration/login.js
```diff
@@ -1,92 +1,118 @@
...
+const clientsPage = new ClientsPage();
+const clientCreatePage = new ClientCreatePage();
+const tasksPage = new TasksPage();
+const taskCreatePage = new TaskCreatePage();
+const taskEditPage = new TaskEditPage();
+const timeLoggingPage = new TimeLoggingPage();
+const common = new Common();
+    
+var roles = [
+    { name: 'User', 'tabCount': 1 },
```
nikolaj-tolkaciov, line 20: first tab count is with quotes, but rest of them are not

cypress/obj/tasks/taskCreatePage.js
```diff
@@ -0,0 +1,27 @@
...
+    getFormName() {
+        return cy.get('[id="taskDetailsForm.name"]');
+    }
+
+    getFormDescription() {
+        return cy.get('[id="taskDetailsForm.description"]');
+    }
+
+    getBillableDropdown() {
+        return cy.get('label').contains('Bill to Client').next();
```
nikolaj-tolkaciov, line 11: is there any other way to get this button instead of by text and next()?

cypress/integration/login.js
```diff
@@ -1,92 +1,118 @@
...
+];
 
-    it('Should display validation for empty user after attempted loggin', function () {
-        
-        cy.visit('/');
-        cy.get('.Select.not-valid').should('not.visible')
-        cy.get('[type="submit"]').click();
-        cy.get('.Select.not-valid').should('be.visible')
-    })
+function makeId() {
```
nikolaj-tolkaciov, line 27: this function should not be in a test case, maybe some kind of utility class with all helping functions would be a good idea to have and keep them there

cypress/integration/login.js
```diff
@@ -1,92 +1,118 @@
...
 
-    function testLogIn(role) {
-        cy.visit('/');
-        cy.get('[id="loginForm.userId"]').click({force:true});
-        cy.get('[aria-label="Marius Lastauskas"]').click();
-        cy.get('[id="loginForm.role"]').click({force:true});
-        cy.get('[aria-label="' + role + '"]').click();
-        cy.get('[type="submit"]').click();
-    }
+function testLogIn(role) {
```
nikolaj-tolkaciov, line 32: this could be a function in loginPage instead

cypress/integration/login.js
```diff
@@ -1,92 +1,118 @@
...
+    loginPage.visit();
+    loginPage.getUserDropdown().click({force:true});
+    loginPage.getUserFromDropdown('Marius Lastauskas').click();
+    loginPage.getRoleDropdown().click({force:true});
+    loginPage.getRoleFromDropdown(role).click();
+    loginPage.getSubmitButton().click();
+}
 
+describe('Sourcebook test Clients', function() {
```
nikolaj-tolkaciov, line 41: This is a test case to create client, but you are still in login.js file, would make sense to put it in another file

cypress/integration/login.js
```diff
@@ -1,92 +1,118 @@
...
-        cy.visit('/clients');
-        cy.get('[class="field--filter"]').first().find('input').type(organization);
-        cy.get('[class="ag-body-viewport-wrapper"').find('.ag-body-viewport').should('have.length', 1);
+        clientsPage.visit();
+        clientsPage.getOrganizationFilter().type(organization);
+        clientsPage.getFilteredList().should('have.length', 1);
     })
+})
 
+describe('Sourcebook test Tasks', function() {
```
nikolaj-tolkaciov, line 61: This is a test case to create task, but you are still in login.js file, would make sense to put it in another file

nikolaj-tolkaciov (changes requested):

linukkis (approved):
Good job 👍 

On second thought, why is your request pointing to T4?

Sorry, I didn't see your response

MariusLastauskas:
Thanks :)

PR #171, opened 2019-08-02
First 3 tasks with object classes

ghost:


cypress/integration/T9.js
```diff
@@ -0,0 +1,87 @@
+import LoginPage from  '../objects/loginPage';
+import Task from '../objects/task';
+var login = new LoginPage();
+var task = new Task();
+
+
+describe('Sourcebooks testLogin', function(){
+    it('T1: TestLogin with Today date', function() {
```
nikolaj-tolkaciov: There is no actual value in keeping any track to T1 in here, you have a branch to do this. but when task is finished, there is no value in knowing it was firs, second or one thousand and fifth, so I would recommend removing this one

cypress/integration/T9.js
```diff
@@ -0,0 +1,87 @@
...
+        login.getPageTitle().contains('Timesheets');
+        login.getCalendar().should('be.visible');
+        login.getTileForm().should('be.visible');
+        login.getUserTitle().contains('Dainius Gaidamavicius');
+        login.getMenuList().should('have.length', 2);          
+        login.getCalendarSpan().should('contain', new Date().getDate());
+    })
+})
+
+describe('Sorcebooks verify all user roles', function() {
```
nikolaj-tolkaciov: this could be a part of previous describe:
describe('Sourcebooks testLogin',

cypress/integration/T9.js
```diff
@@ -0,0 +1,87 @@
...
+        login.getSubmitButton().click();
+        login.getUserTitle().contains('Dainius Gaidamavicius');
+        login.getMenuList().should('have.length', Roles[i].tabs);
+        login.getActiveMenuItem().should('have.css','color', 'rgb(64, 76, 237)');         
+        })
+    } 
+
+})
+
+describe('Sourcebooks creates task', function() {
```
nikolaj-tolkaciov: it would be a good idea to move this to new file, like integration/task.js

cypress/objects/loginPage.js
```diff
@@ -0,0 +1,60 @@
...
+    visit() {
+        cy.visit('/');
+    }
+
+    getUserTitle() {
+        return cy.get('.user-info__title');
+    }
+
+    getURL() {
+        return cy.url();
```
nikolaj-tolkaciov: this function is kind of redundant, it just calls another function with no parameters

cypress/objects/task.js
```diff
@@ -0,0 +1,47 @@
...
+    getTaskRateForm() {
+        return cy.get('[id = "taskDetailsForm.rate"]');
+    }
+
+    getTaskSubmitButton() {
+        return cy.get('[type="submit"]');
+    }
+
+    getTaskNameFieldForFiltering() {
+        return cy.get('[class = "field--filter"]').first().find('input');
```
nikolaj-tolkaciov: you can use y.get('.field--filter') for same results but save some space

cypress/objects/task.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    getViewPortWrapper() {
+       return cy.get('[class="ag-body-viewport-wrapper"').find('.ag-body-viewport')
```
nikolaj-tolkaciov: same as above

cypress/objects/task.js
```diff
@@ -0,0 +1,47 @@
+class task {
+    visitTasks() {
+        cy.visit('/tasks');
+    }
+    
+    getCreateTaskButton() {
+        return cy.get('[type="button"]');
+    }
+
+    getTaskName() {
```
nikolaj-tolkaciov: this is a bit confusing, but maybe try adding get**Random**TaskName this would bring more clarity to the functions purpose and actual behaviour

nikolaj-tolkaciov (changes requested):

cypress/integration/T9.js
linukkis: I think you forgot to remove T3 from this file :)

DavidV44 (approved):

cypress/integration/T3-Task.js
```diff
@@ -0,0 +1,29 @@
+import LoginPage from  '../objects/loginPage';
+import Task from '../objects/task';
+var login = new LoginPage();
+var task = new Task();
+
+describe('Sourcebooks creates task', function() {
+      it('T-3: Admin creates new task', function () {
+        task.visitTasks();
+        login.userDropDown().click({force:true});
```
mantasnakt, line 9: command of LoginAs wasn't added here.

cypress/integration/T3-Task.js
```diff
@@ -0,0 +1,29 @@
...
+        login.getUser("Dainius Gaidamavičius").click();
+        login.getRoleDropDown().click({force:true});
+        login.getRoleOption("Admin").click();
+        login.getSubmitButton().click();
+        task.getCreateTaskButton().click();
+        let taskName = task.getRandomTaskName();
```
mantasnakt, line 15: Randomization of the task name could be done straight in the test. It's not really related to the POM, so moving it away is not required.

cypress/integration/T3-Task.js
```diff
@@ -0,0 +1,29 @@
...
+        task.getTaskNameForm().type(taskName);
+        task.getTaskDescriptionForm().type('TASK11');
+        task.getTaskBillDrop().next().click();
+        task.inputTaskBillOption("Yes").click();
```
mantasnakt, line 19: A perfect way to select options from the dropdowns

cypress/integration/T3-Task.js
```diff
@@ -0,0 +1,29 @@
...
+        task.getTaskRateForm().clear();
+        task.getTaskRateForm().type(Math.random() * 5);
+        task.getTaskSubmitButton().click();
```
mantasnakt, line 22: Just a suggestion to move it away to the class of the common objects.

cypress/integration/T3-Task.js
```diff
@@ -0,0 +1,29 @@
...
+        cy.url().should('not.contain', 'create');
+        task.visitTasks()
+        task.getTaskNameFieldForFiltering().type(taskName);
+        task.getViewPortWrapper().should('have.length', 1);
```
mantasnakt, line 26: The test could be better. Now we don't now if the task that we have found is really the one we created. It could be our task name+some other random symbols and test will pass.
Example:
cy.get('.ellipsis' + [title="${name}"])
Geeting a cell with its title which corresponds to task name we randomized.

mantasnakt (commented):
Great job, however, I`ve noticed the absence of LoginAs command.

PR #172, opened 2019-08-02
Mindaugas maceika/feature/t9 feature branch for page object m refactoring

mindaugasmc:
T9 feature branch refactoring for page object model

### DIFF
--- a/cypress/objects/timeLoggingPage.js
+++ b/cypress/objects/timeLoggingPage.js
@@ -19,11 +19,11 @@ class TimeLoggingPage {
     return cy.get('.tile.form');
   }
 
-  getUserInfoTitle() {
+  getUserName() {
     return cy.get('.user-info__title');
   }
 
-  getMainNav() {
+  getNavigationBar() {
     return cy.get('.main-nav');
   }
 
@@ -31,7 +31,7 @@ class TimeLoggingPage {
     return cy.get('.calendar--today.calendar--selected');
   }
 
-  getMainNavActiveLink() {
+  getActiveTab() {
     return cy.get('.main-nav__link--active');
   }
 }
--- a/cypress/objects/timeLoggingPage.js
+++ b/cypress/objects/timeLoggingPage.js
@@ -1,0 +1,39 @@
+class TimeLoggingPage {
+  visit() {
+    cy.visit('/time-logging');
+  }
+
+  getUrl() {
+    return cy.url();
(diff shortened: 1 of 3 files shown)


cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,39 @@
...
+
+  getCalendar() {
+    return cy.get('.calendar');
+  }
+
+  getTileForm() {
+    return cy.get('.tile.form');
+  }
+
+  getUserInfoTitle() {
```
linukkis: I think something like `getUserName()` or similar would suit better, don't you agree? :) Because even though the name hides under that 'info title' class, it's not exactly clear what 'info title' means and might bring a little confusion when you see `getUserInfoTitle()` somewhere in your code :)

cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,39 @@
...
+    return cy.get('.user-info__title');
+  }
+
+  getMainNav() {
```
linukkis: Same as above, one might ask: what is a 'main nav'? Instead we could name it `getNavigationBar()` or something similar, so it's clear what the function refers to :)

cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,39 @@
...
+    return cy.get('.main-nav');
+  }
+
+  getSelectedDateToday() {
+    return cy.get('.calendar--today.calendar--selected');
+  }
+
+  getMainNavActiveLink() {
```
linukkis: I suppose this refers to the active selected tab? If I'm correct, why not call it something like 'getActiveTab()'? It will make your life easier when you see this function somewhere later on :)

cypress/objects/timeLoggingPage.js
mindaugasmc: Yes, totally agree, thanks :) Already fixed.

cypress/objects/timeLoggingPage.js
mindaugasmc: Fixed :)

cypress/objects/timeLoggingPage.js
mindaugasmc: Fixed :)

linukkis (approved):
Well done! 👍

PR #175, opened 2019-08-02
refactored login and T2

GretaAl:


cypress/objects/timeLogging.js
```diff
@@ -0,0 +1,28 @@
...
+    getPageTitle(){
+        return cy.get('.page__title')
+    }
+    getCalendar(){
+        return cy.get('.calendar')
+    }
+    getTileForm(){
+        return cy.get('.tile.form')
+    }
+    getUserInfoTitles(){
```
DavidV44: This and below methods could be put in `common.js` or `pageComponents.js` files, because this is shared between pages and if you would want to check username it would look weird if you have to use `timeLogging` to do so.

DavidV44 (changes requested):

DavidV44 (approved):
very nice 😎

PR #176, opened 2019-08-02
refactoring t1 and t2

aistyska:


cypress/objects/timeLogging.js
```diff
@@ -0,0 +1,38 @@
...
+        let today = new Date()
+        return cy.get('.calendar--today').find('.calendar__date').contains(today.getDate());
+    }
+
+    getTheColor() {
+        return cy.get('.main-nav__link--active').should('have.css', 'color').and('equal', 'rgb(64, 76, 237)');
+    }
+}
+
+
```
linukkis: There's no need to have 3 empty lines, could you remove 2 and leave just 1? :)

cypress/objects/timeLogging.js
```diff
@@ -0,0 +1,38 @@
...
+    getTabsCount() {
+        return cy.get('.main-nav').find('li')
+    }
+
+    assertTodayDate() {
+        let today = new Date()
+        return cy.get('.calendar--today').find('.calendar__date').contains(today.getDate());
+    }
+
+    getTheColor() {
```
linukkis: Great function, but "get the color" of what? Let's specify that in the function name and call it something like `getActiveTabColor()` :)

cypress/objects/timeLogging.js
```diff
@@ -0,0 +1,38 @@
...
+
+    getCalendar() {
+        return cy.get('.calendar')
+    }
+
+    getTitleForm() {
+        return cy.get('.tile.form')
+    }
+
+    getUserInfoTitles() {
```
linukkis: What is a 'user info title'? What hides behind this class is the name of the user, right? So why not call it something like `getUserName()`? :)

cypress/objects/timeLogging.js
```diff
@@ -0,0 +1,38 @@
...
+
+    getPageTitle() {
+        return cy.get('.page__title')
+    }
+
+    getCalendar() {
+        return cy.get('.calendar')
+    }
+
+    getTitleForm() {
```
linukkis: It's 'tile', not 'title' ;)

cypress/objects/loginPage.js
```diff
@@ -0,0 +1,31 @@
...
+    
+    getSpecificUserFromDropDown(userName) {
+        return cy.get(`[aria-label="${userName}"]`)
+    }
+
+    getLoginFieldRole() {
+        return  cy.get('[id="loginForm.role"]')
+    }
+
+    getUserRole(role) {
```
linukkis: You called the other function `getSpecificUserFromDropDown()`, so this could be `getSpecificRoleFromDropdown()`, right? ;)

DavidV44 (approved):
noice

PR #177, opened 2019-08-02
refactored T3

kasteckis:


cypress/obj/loginPage.js
```diff
@@ -4,6 +4,12 @@ class LoginPage
     {
         cy.visit('/');
     }
+
+    visitSpecificURL(url)
```
linukkis: The way you wrote this function makes it common, and you could extract this to some common.js file and user it from everywhere :)

cypress/integration/t3.js
```diff
@@ -0,0 +1,35 @@
...
+        let taskName = Math.random();
+        let taskDescription = Math.random();
+        let rateValue = Math.floor(Math.random() * 10) + 1;
+
+        loginPage.visit();
+        loginPage.loginAsSpecificUserAndRole("Valentinas Kasteckis", "Admin");
+        loginPage.visitSpecificURL('/tasks');
+
+        t3page.typeInTaskName(taskName);
+        cy.wait(3000); // for some reason JS does not have enough time to load in, so this helps
```
linukkis, line 22: I understand what you're saying. It's fine to keep it here I think, but looking at the long run it's quite bad practice to use wait/sleep in your tests because it's not robust at all, and depending on the internet connection the given milliseconds may not be enough. It's usually best to wait for some element to become visible/invisible instead of putting a thread to sleep. In our system I think you could try waiting for the 'loading' message to disappear, but you can leave this 'wait' for the time being.

linukkis (approved):
Great job!

PR #178, opened 2019-08-02
t3 and previous test updates

JelenaCernysova:


cypress/integration/t-3.js
```diff
@@ -0,0 +1,26 @@
...
+
+const loginPage = new LoginPage();
+const tasksPage = new TasksPage();
+const createTaskPage = new CreateTaskPage();
+
+describe('Create task', function() {
+
+    beforeEach(function(){
+        cy.loginAs("Admin")
+        cy.visit('/')
```
DavidV44: you could go to /tasks directly, in that case it would be less loading times, because now you go to home page and only then to /tasks.

cypress/integration/login.js
```diff
@@ -22,11 +20,11 @@ describe('Sourcebooks login', function() {
         loginPage.loginWithUserAndRole("Jelena Černyšova", "Team Lead")
 
         cy.url().should('include', '/time-logging');
-        pageComponents.getPageTitle().contains('Timesheets')
+        timeLoggingPage.getPageTitle().contains('Timesheets')
         timeLoggingPage.getCalendar().should('be.visible')
         timeLoggingPage.getTileForm().should('be.visible')
-        pageComponents.getLoggedInUser().contains('Jelena Cernyšova');
-        pageComponents.getNavigationTabsList().should('have.length', 2)
+        timeLoggingPage.getLoggedInUser().contains('Jelena Cernyšova');
```
DavidV44, line 26: Why did you change back to using `timeLoggingPage` ? Using `pageComponents` seemed better

DavidV44 (changes requested):

cypress/integration/login.js
JelenaCernysova: it uses pageComponents, as TimeLoggingPage class extends pageComponents  class, therefore the logic of pageComponents is in TimeLoggingPage  class without creating additional variables in a test and the logic can still be used

DavidV44: Oh, didn't see that, all good then :)

DavidV44 (approved):

PR #181, opened 2019-08-02
Ieva stonkaite/feature/t 9 feature branch for pom refactoring

ghost:
Adjusted the branch so that the the pull request points to my local develop at ieva_stonkaite/develop


cypress/objects/t2Page.js
```diff
@@ -0,0 +1,66 @@
...
+    getInsertValueIntoTaskDescrField(taskDescr) {
+        return cy.get('[id="taskDetailsForm.description"]').type(taskDescr);
+    }
+
+    getBillToClientDrowpDown() {
+        return cy.get('div.Select-value').click();
+    }
+
+    
+
```
DavidV44: a bit too much of empty lines :)

cypress/objects/t2Page.js
```diff
@@ -0,0 +1,66 @@
+class T2Page {
```
DavidV44: This file should be name `timeLoggingPage `or something like that, because T2 doesn't say much :)

cypress/objects/t2Page.js
```diff
@@ -0,0 +1,66 @@
...
+    visit() {
+        cy.visit('/'); 
+    }
+
+    getUserDropDown() {
```
DavidV44: This and some of other below methods are are in `loginPage.js` file. This file should only contain methods that work with elements of time logging page

cypress/integration/t2.js
```diff
@@ -0,0 +1,73 @@
+import T2Page from '../objects/t2Page';
+
+const t2Page = new T2Page();
```
DavidV44: you should name this `timeLoggingPage`, T2 is a bad name. Also you should import Login page and use those methods for login

DavidV44 (changes requested):

cypress/integration/t2.js
```diff
@@ -0,0 +1,68 @@
+import LoginPage from '../objects/loginPage';
+
+const loginPage = new LoginPage();
+
+it('Should verify all user roles can log in and should see appropriate tabs - USER', function () {
```
DavidV44: I think these test should be in `login.js` file as they check login functionality

cypress/objects/loginPage.js
```diff
@@ -0,0 +1,42 @@
...
+
+    getRoleDropDown() {
+        return cy.get('[id="loginForm.role"]')
+    }
+
+    getSpecificRoleFromDropDown(role) {
+        return cy.get(`[aria-label="${role}"]`)
+    }
+    
+    getUserTitle() {
```
DavidV44, line 30: Methods like these could be put in `common.js` or` pageComponents.js `files, because this is shared between pages and if you would want to check username or navigation bar it would look weird if you have to use `loginPage `to do so.

cypress/objects/timeLoggingPage.js
```diff
@@ -0,0 +1,24 @@
...
+
+    getCalendarTodaysDate() {
+        return cy.get('.calendar--selected').find('span').contains(new Date().getDate())
+    }
+
+    getPageTitle() {
+        return cy.get('.page__title')
+    }
+    
+    getUserTitle() {
```
DavidV44: This also should be put in some kind of shared file like `common.js`, see above comment

DavidV44 (changes requested):

cypress/integration/login.js
```diff
@@ -1,103 +1,110 @@
...
+        common.getSpecificRoleFromDropDown("Team Lead").click();
+        loginPage.getSubmitButton().click();
 
-        cy.url().should('include', '/time-logging');
-        cy.get('.page__title').contains('Timesheets')
-        cy.get('.calendar').should('be.visible')
-        cy.get('.tile.form').should('be.visible')
-        cy.get('.user-info__title').contains('Ieva Stonkaite');
-        cy.get('.main-nav').find('li').should('have.length', 2);
+        cy.wait(2000); // for reason js sometimes doesn't load my client
```
nikolaj-tolkaciov, line 26: what is this about? it seems it work for others, try switching places with  common.getPageTitle().contains('Timesheets');

then it would wait for page title to change and afterwards assert URL, should be more stable that way, as URL changes during page loading and this might throw off cypress assert

cypress/integration/login.js
```diff
@@ -1,103 +1,110 @@
...
+        common.getCalendar().should('be.visible');
+        common.getPageTitle().should('be.visible');
+        common.getUserTitle().contains('Ieva Stonkaite');
+        loginPage.getElementCountOfNavigationBar(2);
     })
 
-    it('Should show calendar date', function () {
-        cy.get('.calendar--selected').find('span').contains(new Date().getDate());
+    it('Should validate what date is selected as “Today” on Time Logging page', function () {
+        common.getCalendarTodaysDate();
```
nikolaj-tolkaciov, line 36: this function name is kind of misleading, because inside it actually assert as well, not only getting you todays date

cypress/objects/common.js
```diff
@@ -0,0 +1,40 @@
+class Common {
+    
+    getSpecificUrl(url) {
+        cy.visit(url); 
+    }
```
nikolaj-tolkaciov, line 5: there is no point of having this function, as it just takes url and pass it to another function, you can call cy.visit(url) with same results

cypress/objects/loginPage.js
nikolaj-tolkaciov: This is not yet fixed

nikolaj-tolkaciov (changes requested):
Found few remarks, but other then that it is almost finished!

PR #184, opened 2019-08-02
t4

JelenaCernysova:


cypress/objects/clientsPage.js
```diff
@@ -0,0 +1,18 @@
...
+        cy.visit('/clients');
+    }
+
+    pressCreateNewButton(){
+        cy.get('[class="btn"]').contains('Create Client').click();
+    }
+
+    assertClientCreated(organization){
+        this.getFirstSearchField().type(organization + '{enter}')
+        cy.get('[class="ag-cell ag-cell-not-inline-editing ag-cell-no-focus ag-cell-value"]').contains('LastName, FirstName')
```
DavidV44: can't this selector be simpler?

JelenaCernysova: I did not find a simpler one to use

DavidV44: `.ag-cell` is a simpler one

JelenaCernysova: I have already changed it to look for the correct title, or is this a better approach?

DavidV44: well, this is a shorter selector that selects the same element as yours. It's worth mentioning that shorter selectors are not always better.

DavidV44: In any case i thunk your selector is ok for this test, but if you ever see a very long selector try to find ways to make it simpler.

DavidV44 (approved):

PR #187, opened 2019-08-02
Dainius gaidamavicius/feature/t4 admin creates client

ghost:


cypress/integration/T9.js
```diff
@@ -0,0 +1,65 @@
...
+            "tabs": 5
+        }, 
+        {
+            "name": "Admin",
+            "tabs": 6
+        }];
+    
+    for(let i = 0; i < Roles.length; i++){
+        it('Verify that ' + Roles[i].name + ' can log in and should see ' + Roles[i].tabs + ' tabs', function(){
+        login.visit();
```
DavidV44, line 50: bad indent

cypress/objects/client.js
```diff
@@ -0,0 +1,56 @@
+class Client {
```
DavidV44, line 1: This file could be called `ClientPage.js` because just "client" doesn't say much

cypress/objects/task.js
```diff
@@ -0,0 +1,47 @@
+class task {
```
DavidV44, line 1: Same as with client, `taskPage.js` would fit as a better file name

cypress/integration/T3-Task.js
```diff
@@ -0,0 +1,29 @@
...
+        task.getTaskDescriptionForm().type('TASK11');
+        task.getTaskBillDrop().next().click();
+        task.inputTaskBillOption("Yes").click();
+        task.getTaskRateForm().clear();
+        task.getTaskRateForm().type(Math.random() * 5);
+        task.getTaskSubmitButton().click();
+        cy.url().should('not.contain', 'create');
+        task.visitTasks()
+        task.getTaskNameFieldForFiltering().type(taskName);
+        task.getViewPortWrapper().should('have.length', 1);
```
DavidV44, line 26: You should also check that task that you just created have same name description and so on, because there is possibility that somehow task with wrong description could be created and the test would still pass

DavidV44 (changes requested):

PR #190, opened 2019-08-02
Modestas kmieliauskas/feature/t3 admin creates new task

Modkmi:


cypress/objects/TaskPage.js
```diff
@@ -1,3 +1,37 @@
...
+    getBillToClientOption() {
+        return cy.get('[aria-label="Yes"]');
+    }
+    getHourlyRateField() {
+        return cy.get('[id="taskDetailsForm.rate"]');
+    }
+    getSubmitButton() {
+        return cy.get('[type="submit"]');
+    }
+    assertUrl() {
```
DavidV44: `assertUrl` doesn't say what kind of url it asserts. I suggest either passing what king of url you want assert, or renaming method to `assertNotCreateUrl` or something like that

cypress/integration/login.js
```diff
@@ -1,44 +1,63 @@
...
+                loginPage.getUserOption(userName).click();
+                loginPage.getRoleDropDown().click({ force: true });
+                loginPage.getRoleOption(roles[i]).click();
+                loginPage.getSubmitButton().click();
+
+                timeLoggingPage.assertUrl();
+                timeLoggingPage.getPageTitle().contains('Timesheets');
+                timeLoggingPage.getCalendarSelector().should('be.visible');
+                timeLoggingPage.getTileForm().should('be.visible');
+                timeLoggingPage.getUserInfoTitle().contains(userName);
```
DavidV44: Methods like these could be put in `common.js` or `pageComponents.js` files, because this is shared between pages and if you would want to check username or navigation bar it would look weird if you have to use `TimeLoggingPage `to do so.

DavidV44 (changes requested):

PR #191, opened 2019-08-02
completed T4

kasteckis:


cypress/obj/t4Page.js
```diff
@@ -0,0 +1,37 @@
+class T4Page
```
DavidV44: Name files with meaningful name, like `clientPage `and so on

cypress/integration/j4.js
```diff
@@ -0,0 +1,39 @@
+import T3Page from '../obj/t3Page';
```
DavidV44: Name files with meaningful name, like `clientPage `and so on

cypress/integration/j4.js
```diff
@@ -0,0 +1,39 @@
...
+import T4Page from '../obj/t4Page';
+import LoginPage from '../obj/loginPage';
+import Common from '../obj/common';
+
+const t3page = new T3Page();
```
DavidV44: same as above

DavidV44 (changes requested):

cypress/integration/j5.js
```diff
@@ -0,0 +1,30 @@
+import T3Page from '../obj/tasksPage';
```
DavidV44, line 1: why is file name `j5.js`, also as a general note i would suggest using names that reflect what kind of test i would find inside file, like `login.js` or `task.js`

cypress/integration/j5.js
```diff
@@ -0,0 +1,30 @@
...
+describe('Sourcebooks login', function() {
+
+    beforeEach(function () {
+        cy.loginAs("Admin");
+        cy.visit('/');
+    })
+
+    it('Admin creates new project', function(){
+
+        /* THIS T4 TASK IS NOT FINISHED */
```
DavidV44, line 20: usually you should not put unfinished functionality in pull request, because it is expected that develop branch is fully working and don't spew errors, but for this lecture it;s okay

DavidV44 (changes requested):

PR #192, opened 2019-08-02
T3

kamilestu:


cypress/integration/login.js
```diff
@@ -88,5 +81,37 @@ describe('Sourcebooks login', function() {
         }
 
         rolesArray.forEach(testFunction);        
+    })    
+})
+
+describe('Sourcebook admin tasks', function() {
```
DavidV44, line 87: You should move this whole task describe block another file like `task.js`, because it looks weird having task tests in `login.js `file next to login test cases

DavidV44 (changes requested):

PR #193, opened 2019-08-02
Admin creates a new task

dominykasposkus:


cypress/objects/common.js
```diff
@@ -1,7 +1,27 @@
...
+        cy.get(`[type="${buttonType}"]`).contains(text).click();
+    }
+
+    generateAUniqueString() {
+        return Date.now().toString();
+    }
+
+    visit(page) {
+        cy.visit(page);
+    }
```
nikolaj-tolkaciov, line 16: this function is redundant, it takes variable and pass it to another function, you could just call that function instead. you can have visit('/tasks') action in taskPage.js for visit action

cypress/objects/common.js
```diff
@@ -1,7 +1,27 @@
...
-        cy.url().should('include', text);
+    checkIfUrlDoesNotInclude(text) {
+        cy.url().should('not.include', text);
     }
+
+    clickCertainButton(buttonType, text) {
+        cy.get(`[type="${buttonType}"]`).contains(text).click();
+    }
+
+    generateAUniqueString() {
```
nikolaj-tolkaciov, line 10: better to call this function getCurrentTimeStamp, as it would be more appropriate name for it

cypress/objects/common.js
```diff
@@ -1,7 +1,27 @@
 class Common {
-    checkIfUrlIncludes(text) {
-        cy.url().should('include', text);
+    checkIfUrlDoesNotInclude(text) {
+        cy.url().should('not.include', text);
     }
+
+    clickCertainButton(buttonType, text) {
```
nikolaj-tolkaciov, line 6: better name would be clickCertainButtonWithText

cypress/objects/taskPage.js
```diff
@@ -0,0 +1,36 @@
...
+    clickBillToClientDropDown() {
+        cy.get('[class = Select-value]').click();
+    }
+
+    selectRandomBillToClientOption(selection) {
+        cy.get(`[aria-label= "${selection}"]`).click();
+    }
+
+    checkIfBillToClientIsYes() {
+        return cy.get('[class="Select has-value is-searchable Select--single"]').get('[taskDetailsForm.billable]').contains(true);
```
nikolaj-tolkaciov, line 24: not sure what you wanted to achieve with this contains(true), maybe .should('be.visible') would be better?

nikolaj-tolkaciov: or if you want to see it contains "Yes"

cypress/integration/T3.js
```diff
@@ -0,0 +1,38 @@
...
+        taskPage.selectRandomBillToClientOption(selection);
+
+        if (selection === 'Yes') {
+            common.clearField('taskDetailsForm.rate');
+            taskPage.inputIntoASpecificField('taskDetailsForm.rate', Math.floor(Math.random()*50));
+        }
+
+        common.clickCertainButton('Submit','Save');
+
+        common.wait(200);
```
nikolaj-tolkaciov, line 31: explicit waits are evil and should be avoided, better off assert for the actual url, or for some kind of page title (page loaded) and then assert for common.checkIfUrlDoesNotInclude('/create');

cypress/integration/T3.js
```diff
@@ -0,0 +1,38 @@
...
+        taskPage.inputIntoASpecificField('taskDetailsForm.name', taskName);
+        taskPage.inputIntoASpecificField('taskDetailsForm.description', Date.now());
+        taskPage.clickBillToClientDropDown();
+        const selection = taskPage.generateRandomBillToClientSelection();
+        taskPage.selectRandomBillToClientOption(selection);
+
+        if (selection === 'Yes') {
+            common.clearField('taskDetailsForm.rate');
+            taskPage.inputIntoASpecificField('taskDetailsForm.rate', Math.floor(Math.random()*50));
+        }
```
nikolaj-tolkaciov, line 27: it is nice approach, but you are gambling here, and there is a chance that you miss a bug. If for example bill to client = yes does not work (have bug) you might run test case n times in a row and it would not show you that there is a bug, and your test would be kind of random

cypress/integration/T3.js
```diff
@@ -0,0 +1,38 @@
...
+    beforeEach(function () {
+        cy.loginAs('Admin');
+        common.visit('/Tasks');
+    })
+
+    it('Admin creates a new task', function () {
+        
+        common.clickCertainButton('button','Create Task');
+        const taskName = Date.now().toString(); 
+        taskPage.inputIntoASpecificField('taskDetailsForm.name', taskName);
```
nikolaj-tolkaciov, line 18: taskDetailsForm.name is an ID of the field and it is stored in a test case, so this approach does not follow page object model. If this field id changes to something else, you would still need to go through each of your test case and change it manually

nikolaj-tolkaciov (changes requested):

PR #194, opened 2019-08-02
T3/HW - Admin creates new task

EdvinasKveselys:


cypress/integration/adminTests.js
```diff
@@ -0,0 +1,30 @@
...
+
+    beforeEach(function(){
+        cy.loginAs("Admin");
+        cy.visit('/');
+    })
+
+    const title = "hello worlds" + Date.now();
+
+    it('Admin creates new task', function() {
+        common.getNavLink("Tasks").click();
```
mantasnakt, line 17: Nice implementation

cypress/integration/adminTests.js
```diff
@@ -0,0 +1,30 @@
...
+        taskPage.createTaskButtonClick();
+        taskPage.getTaskDetailsFormName().click().type(title);
+        taskPage.getTaskDetailsFormDescription().click().type('hello worlds description');
+        taskPage.getTaskBillableDropDown().click({ force: true });
+        taskPage.getYesFromDropDown().click();
+        taskPage.getTaskDetailsFormRate().click().clear().type('1');
+        common.getSubmitButton().click();
+        common.getNavLink("Tasks").click();
+        common.urlShouldNotInclude("/create")
```
mantasnakt, line 26: I would suggest using only one method for checking what's included in URL and then just defining whether it should be included or not.
`url().should('not.contain', 'create');`
However it`s not a mistake, both ways are god, but in one method case, we can shorten our code by 3 lines :)

cypress/objects/taskPage.js
```diff
@@ -0,0 +1,36 @@
...
+
+    getTaskDetailsFormDescription(){
+        return cy.get('[id="taskDetailsForm.description"]')
+    }
+
+    getTaskBillableDropDown(){
+        return cy.get('[aria-selected="true"]')
+    }
+
+    getYesFromDropDown(){
```
mantasnakt, line 18: this one could be parameterized in order to be able to select both - Yes and No in one method.

cypress/objects/taskPage.js
```diff
@@ -0,0 +1,36 @@
...
+        return cy.get('[aria-label="Yes"]')
+    }
+
+    getTaskDetailsFormRate(){
+        return cy.get('[id="taskDetailsForm.rate"]')
+    }
+
+    getTaskNameField(){
```
mantasnakt, line 26: The name field has its own ID as well.

mantasnakt (commented):
Great job! the common class was introduced for common selectors. Mistakes are trivial and code quality is overall good.

PR #195, opened 2019-08-02
t5

JelenaCernysova:


DavidV44 (approved):
Everything looks neat and dandy, but i just noticed that you use T-5 and so on for test file names. In the future i would suggest using names that reflect what kind of test  i would find inside file, like `login.js` or `task.js`

PR #196, opened 2019-08-02
T4 admin creates new client

dominykasposkus:


cypress/integration/T4.js
```diff
@@ -0,0 +1,30 @@
...
+        common.inputIntoASpecificField('clientDetailsForm.contacts_lastName_0', uniqueString);
+        common.inputIntoASpecificField('clientDetailsForm.contacts_homePhone_0', uniqueString);
+        common.inputIntoASpecificField('clientDetailsForm.contacts_mobilePhone_0', uniqueString);
+        common.inputIntoASpecificField('clientDetailsForm.contacts_email_0', 'testingemail@testing.com')
+        common.clickCertainButton('Submit','Save');
+        common.checkIfUrlDoesNotInclude('/clients/create');
+
+        common.visit('/clients');
+        common.filterListItemsByName(uniqueString);
+        common.checkIfListContains(uniqueString);
```
DavidV44, line 27: You should also check that client that you just created have same contacts and so on, because there is possibility that somehow a client with wrong contacts could be created and the test would still pass

DavidV44 (changes requested):

PR #198, opened 2019-08-02
T3 admin is able to create a new task

rasakap:
T3 as HW with new object class 'tasks' and changed ID with name + surname on 'commands'

### DIFF
--- a/cypress/integration/new_task_creation_admin.js
+++ b/cypress/integration/new_task_creation_admin.js
@@ -19,14 +19,14 @@ describe('Sourcebooks task creation', function() {
     it('Admin should create a new task', function () {
 
         const taskName = "Task name " + Math.ceil((Math.random())*10000); 
-        const descriptionName = "Description name " + Math.ceil((Math.random())*10000);       
+        const descriptionText = "Description name " + Math.ceil((Math.random())*10000);       
 
         common.getMainNavigationBar().find('[href="/tasks"]').click(); 
-        common.geturl().should('include', '/tasks');
+        cy.url().should('include', '/tasks');
         common.getPageTitle().contains('Tasks');
         tasks.getCreateTaskButton().click();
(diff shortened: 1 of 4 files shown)


cypress/integration/new_task_creation_admin.js
```diff
@@ -0,0 +1,41 @@
+import Tasks from '../objects/tasks';
```
DavidV44, line 1: I suggest naming files in camel case, like `newTaskCreationAdmin.js` or simply name this file  `newTaskCreation.js`

cypress/integration/new_task_creation_admin.js
```diff
@@ -0,0 +1,41 @@
...
+        tasks.getSelectionValueYes().click({ force: true });
+        tasks.getRateFromTaskDetailsForm().clear().type("0.250");
+        tasks.getSaveButton().click({force:true});
+
+        cy.wait(1000);
+        cy.url().should('to.not.equal', ('/tasks/create'));
+
+        tasks.visit();
+        tasks.getSearchFieldInputField().first().click().type(taskName);
+        cy.get('[title=' + taskName + ']').should('exist');
```
DavidV44: You should also check that task that you just created have same name description and so on, because there is possibility that somehow a task with wrong description could be created and the test would still pass

cypress/objects/tasks.js
```diff
@@ -0,0 +1,46 @@
...
+
+    getSearchFieldInputField() {
+        return cy.get('.field__text')
+    }
+
+    getTaskNamebyTaskId() {
+        return cy.get(`[title=' + "${taskName}" + ']`)
+    }
+
+
```
DavidV44: too much empty lines

DavidV44 (changes requested):

PR #199, opened 2019-08-02
Kamile stugyte/feature/t4 admin creates new client

kamilestu:


cypress/integration/admin.js
```diff
@@ -0,0 +1,71 @@
...
+        general.getSpecificValueFromDropDown('Yes').click();
+        createTasks.getField('rate').clear().type(2);
+        general.getSubmitButton().click();
+
+        //Check if task was created
+        general.getUrl().should('not.include', 'create');
+        general.getItemFromMenu('tasks').click();
+        general.getFilterFields().first().type(taskName);
+        general.getDisplayedItems().contains(taskName);
+    })*/
```
nikolaj-tolkaciov, line 47: There is two commented you test cases. these should be deleted or un-commented in final version

cypress/integration/admin.js
```diff
@@ -0,0 +1,71 @@
...
+
+    it('Admin creates new client', function() {
+
+        //Go to clients page
+        general.getItemFromMenu('clients').click();
+        general.getRegularButton().click();
```
nikolaj-tolkaciov, line 53: this is very random, and it would be better to at least find button by text

cypress/integration/admin.js
```diff
@@ -0,0 +1,71 @@
...
+
+        //Create unique organization name
+        let orgName = 'TestOrg' + new Date().getTime();
+
+        //Fill form fields and submit
+        createClient.getField('organization').type(orgName);
```
nikolaj-tolkaciov, line 59: you are still using locators in the test case, this is not a recommended approach from page object model, all locators should be in page object file and not encoded in the test case file

cypress/integration/admin.js
```diff
@@ -0,0 +1,71 @@
+import LoginPage from '../objects/loginPage';
+import General from '../objects/general';
+import TimeLogging from '../objects/timeLogging';
+import CreateTasks from '../objects/createTasks';
+import CreateClient from '../objects/createClient';
+
+const loginPage = new LoginPage();
+const general = new General();
+const timeLogging = new TimeLogging();
```
nikolaj-tolkaciov, line 9: this is not used in test case

cypress/integration/admin.js
```diff
@@ -0,0 +1,71 @@
...
+const createTasks = new CreateTasks();
```
nikolaj-tolkaciov, line 10: this is not used in test case

nikolaj-tolkaciov (changes requested):
Almost there! only few remarks left

PR #200, opened 2019-08-02
t6

JelenaCernysova:


cypress/objects/timeLoggingPage.js
```diff
@@ -33,8 +33,8 @@ class TimeLoggingPage extends PageComponents{
     }
 
     assertTimeEntrySaved(description){
-        this.getElementByClass("time-list__description").contains(description)
-        this.getElementByClass("time-list__hours").contains("1.00")
+        cy.get(`[title="${description}"]`)
+        this.getElementByClass("time-list__hours").eq(-1).contains("1.00")
```
DavidV44: I suggest changing time logged to 0.1, because if you run this 24 times a day the test would fail (because you tried to log more than 24 hours). Running a test 240 time per day has much lower probability :)

DavidV44 (changes requested):

nikolaj-tolkaciov (approved):

PR #211, opened 2019-08-05
Finished homework

aistyska:


cypress/objects/creatTask.js
```diff
@@ -0,0 +1,23 @@
+class CreatTask {
```
mantasnakt, line 1: CreateTask could live together with the TaskPage objects as those screens represent same page - Tasks CRUD

cypress/integration/Homework_T3_Admin_creates_tasks.js
```diff
@@ -0,0 +1,40 @@
...
+        creatTask.getHourlyRateField().click().clear().type((Math.random() * 100).toFixed(3))
+        pageComponents.getSubmitButton().click()
+
+        cy.wait(1000)
+        cy.url().should('not.include', '/create') 
+
+        taskPage.visit()
+        taskPage.getTaskNameField().click().type(title)
+        taskPage.getDescriptionField().click().type(taskDescription)
+        expect(cy.get(`[title=${title}]`)).to.exist
```
mantasnakt, line 36: Could be moved to separate components as well. "title" name could be a subject of change.

cypress/objects/creatTask.js
```diff
@@ -0,0 +1,23 @@
...
+    getTaskNameField() {
+        return cy.get('[id="taskDetailsForm.name"]')
+    }
+
+    getDescriptionField() {
+        return cy.get('[id="taskDetailsForm.description"]')
+    }
+
+    getBillToClientSelect() {
```
mantasnakt, line 10: It's not clear what Select means. Would be better to include field type into the method name, i.e. "dropdown"

cypress/objects/taskPage.js
```diff
@@ -0,0 +1,19 @@
+class TaskPage {
+    getCreatTaskButton() {
+        return cy.get('.btn[type="button"]')
+    }
+
+    getTaskNameField() {
```
mantasnakt, line 6: suggest using id`s instead of class names if possible - cy.get('[id="taskDetailsForm.name"]')

cypress/integration/Homework_T3_Admin_creates_tasks.js
```diff
@@ -0,0 +1,40 @@
...
+        
+        taskPage.getCreatTaskButton().click()
+        creatTask.getTaskNameField().click().type(title)
+        creatTask.getDescriptionField().click().type(taskDescription)
+        creatTask.getBillToClientSelect().click()
+        creatTask.getBillToClientYesOption().click()
+        creatTask.getHourlyRateField().click().clear().type((Math.random() * 100).toFixed(3))
+        pageComponents.getSubmitButton().click()
+
+        cy.wait(1000)
```
mantasnakt, line 30: this part should be working without waiting

mantasnakt (commented):
Overall - a great job! page object model used at its best. Even a common component class was introduced.

PR #231, opened 2019-08-27
Configured for devbstaging lunch website. Added automated login, dish…

Modkmi:
… selection and validation.


EdvinasKveselys (changes requested):
automatedTests.js name should be more specific according to the test case

